PR #1, opened 2018-06-06
Add npm-watch for auto-rebuilding percy-agent-js.

fotinakis:


anaulin:
@fotinakis  @djones any reason why this PR ended up not being merged?

If there is no objection against it, I'd like to add a slightly broader version of this PR, to rebuild everything on any Typescript change. I can either update this PR or make a new one -- let me know what you prefer.

package.json
```diff
@@ -6,6 +6,16 @@
   "bin": {
     "percy": "./bin/run"
   },
+  "watch": {
+    "build": {
+      "patterns": [
+        "src"
```
djones: I think this should be limited to just `/src/percy-agent-client`

The rest of the source code does not need to build because `oclif` uses `ts-node` which is able to execute TypeScript directly.

More info over at https://oclif.io/docs/faqs#should-i-use-typescript-or-javascript

anaulin: Ah, tricky. Fixed.

PR #5, opened 2018-09-12
Allow PercyAgent to be exported onto the window as PercyAgent

djones:
screen shot 2018-09-12 at 10 30 45 am


djones:
This comment was helpful at figuring this out: https://github.com/TypeStrong/tsify/issues/105#issuecomment-241936617

src/percy-agent-client/percy-agent-service.ts
```diff
@@ -0,0 +1,13 @@
+export class PercyAgentService {
```
fotinakis: This is more like a `PercyAgentClient`, right?

src/percy-agent-client/percy-agent.ts
```diff
@@ -37,7 +19,7 @@ export class PercyAgentClient {
   snapshot(name: string, options: SnapshotOptions = {}) {
     let documentObject = options.document || document
     let domSnapshot = this.domSnapshot(documentObject)
-    let percyAgent = new PercyAgent(this.xhr)
+    let percyAgent = new PercyAgentService(this.xhr)
```
fotinakis: Maybe `percyAgentService`?

src/percy-agent-client/percy-agent-service.ts
fotinakis: He says, non-ironically.

PR #8, opened 2018-11-01
Introduce .percy.yml configuration file for setting snapshot defaults

djones:
This PR allows the user to create a `.percy.yml` file like this:

```yaml
version: 1
snapshot:
  widths: [375, 1280]
  min-height: 1024
  enable-javascript: false
```

This configuration will be picked up as the default options when taking a snapshot. This is particularly useful for setting default snapshot widths across for your entire build.

The configuration file needs to be created in the directory the `percy` command is being run. For now, `.percy.yml` is hardcoded as the filename. Potentially in the future we'll have a flag which will allow you to point to any file on disk to be used for the configuration.

### DIFF
--- a/src/utils/configuration.ts
+++ b/src/utils/configuration.ts
@@ -5,7 +5,6 @@ let path = require('path')
 export interface SnapshotConfiguration {
   widths?: [number],
   'min-height'?: number,
-  'enable-javascript'?: boolean
 }
 
 export interface Configuration {
--- a/src/utils/configuration.ts
+++ b/src/utils/configuration.ts
@@ -13,10 +13,10 @@ export interface Configuration {
 }
 
 let configuration = (relativePath = '.percy.yml'): Configuration => {
-  const confFilePath = path.join(process.cwd(), relativePath)
+  const configFilePath = path.join(process.cwd(), relativePath)
 
   try {
-    return yaml.safeLoad(fs.readFileSync(confFilePath, 'utf8'))
+    return yaml.safeLoad(fs.readFileSync(configFilePath, 'utf8'))
   } catch {
     // this is ok because we just use this configuration as one of the fallbacks
     // in a chain. snapshot specific options -> agent configuration -> default values
--- a/src/utils/configuration.ts
+++ b/src/utils/configuration.ts
@@ -1,0 +1,17 @@
+let yaml = require('js-yaml')
+let fs = require('fs')
+import logger from './logger'
+
+let configuration = (): any => {
+  let conf = {}
+
+  try {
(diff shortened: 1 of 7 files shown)


src/utils/configuration.ts
```diff
@@ -0,0 +1,28 @@
+let yaml = require('js-yaml')
+let fs = require('fs')
+let path = require('path')
+
+export interface SnapshotConfiguration {
+  widths?: [number],
+  'min-height'?: number,
+  'enable-javascript'?: boolean
```
fotinakis: I thought we decided not to include this option to reduce confusion?

djones: Gone.

src/utils/configuration.ts
```diff
@@ -0,0 +1,27 @@
...
+  'min-height'?: number,
+}
+
+export interface Configuration {
+  version: number,
+  snapshot: SnapshotConfiguration
+}
+
+let configuration = (relativePath = '.percy.yml'): Configuration => {
+  const confFilePath = path.join(process.cwd(), relativePath)
```
fotinakis:
```suggestion
  const configFilePath = path.join(process.cwd(), relativePath)
```

src/utils/configuration.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  try {
+    return yaml.safeLoad(fs.readFileSync(confFilePath, 'utf8'))
```
fotinakis:
```suggestion
    return yaml.safeLoad(fs.readFileSync(configFilePath, 'utf8'))
```

test/utils/configuration.test.ts
```diff
@@ -0,0 +1,18 @@
+import {expect} from 'chai'
+import configuration from '../../src/utils/configuration'
+
+describe('Configuration', () => {
+  it('parses valid configuration', () => {
```
fotinakis: What about an invalid configuration format?

src/utils/configuration.ts
djones: Accepted with dc82320fa7d7c7329165cfd41eb0af1fdd84df43

test/utils/configuration.test.ts
djones: I've realized invalid configuration probably doesn't matter all that much in the current context because all supported values have good default fallbacks.

I had an initial idea to use TypeScript to validate the shape of the loaded configuration, but that's not actually possible because typings are checked at compile not run time. There are some schema libraries out there we could use, but given we have fallbacks for all values you can use right now I'm not sure it's all that important.

As we build out our configuration more, we might want to consider `percy config validate` to allow people to check their files.

PR #11, opened 2018-11-09
Bump to puppeteer 1.10.0 

djones:
Patches this issue https://github.com/GoogleChrome/puppeteer/issues/2888 that we have been receiving.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "express": "^4.16.3",
     "js-yaml": "^3.12.0",
     "percy-client": "^3.0.1",
-    "puppeteer": "^1.7.0",
+    "puppeteer": "^1.10.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "express": "^4.16.3",
     "js-yaml": "^3.12.0",
     "percy-client": "^3.0.1",
-    "puppeteer": "^1.7.0",
+    "puppeteer": "^1.10.0",
     "retry-axios": "^0.3.2",
     "sinon": "^6.1.5",
```
Robdel12: Unrelated, but should this be a dependency?

djones: Nope! Feel free to issue a PR for that @Robdel12. Good find.

PR #21, opened 2018-11-20
Adopt @percy/tslint

djones:
Adopt shared TypeScript linting from https://www.npmjs.com/package/@percy/tslint


src/services/build-service.ts
```diff
@@ -18,32 +18,6 @@ export default class BuildService extends PercyClientService {
...
-  // ;@##############WW5#########                  ,#W^                \n\
-  //      |@######"     @######^                ;#W                    \n\
-  //     ;@#####b      ;####"                 @##C                     \n\
-  //    ;######b                            ;@##S                      \n\
-  //   ;######C                            ;####O                      \n\
-  //  |5"\'@@##         ;@####"6#p         ;@####C                      \n\
-  //      |@#S       @#####U    \'6p      .@#####p                      \n\
-  //       9#S     ;#\'@###U       |S     @\'\'5###                       \n\
-  //        7#p  ;S\'   77\'         \'@p  $p   75"                       \n\
-  //          """                    \'""'))
```
Robdel12: 😢 RIP lil percy 🦔

djones: Yeah might add it back later (but something smaller).

test/support/chai.ts
```diff
@@ -1,6 +1,8 @@
 import * as chai from 'chai'
+import * as sinonChai from 'sinon-chai'
 
+// tslint:disable-next-line:no-var-requires
 chai.use(require('chai-http'))
```
djones: I wasn't able to find a way not to use `require` here.

src/commands/exec.ts
```diff
@@ -9,19 +9,19 @@ export default class Exec extends PercyCommand {
 
   static examples = [
     '$ percy exec -- echo \"percy is running around this echo command\"',
-    '$ percy exec -- bash -c "echo foo && echo bar"'
+    '$ percy exec -- bash -c "echo foo && echo bar"',
```
djones: I've gone with requiring a comma on the last items in arrays etc. This matches a decision we made for the Ruby codebase. It allows you to swap lines around easily without having to manage commas.

src/percy-agent-client/index.ts
```diff
@@ -1,3 +1,5 @@
 // This is setup like this so you can include percy-agent.js onto your webpage
 // and then simply use `var percyAgent = new PercyAgent()` to create a new client instance.
+
+// tslint:disable-next-line:no-var-requires
 module.exports = require('./percy-agent').default
```
djones: This is a particular special line that I'd prefer to keep in this style.

bin/run
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
 
+// tslint:disable:no-var-requires no-submodule-imports
 require('@oclif/command').run()
 .catch(require('@oclif/errors/handle'))
```
djones: This one is a bit of a special case. Something provided to us via Oclif.

Issue #30, opened 2018-12-02
"Synchronous XMLHttpRequest on the main thread is deprecated" warnings


toolmantim:
When using https://github.com/percy/percy-puppeteer I receive the following warning:

```
(node:72) UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: console.warning not allowed in production: Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.
```

It looks like it's due to https://github.com/percy/percy-agent/blob/master/src/percy-agent-client/percy-agent-client.ts#L18

djones:
Hi @toolmantim,

I believe this is now fixed, especially for the `@percy/puppeteer` SDK because we've changed the way a page communicates with the agent.

It used to communicate directly from within the page being tested (and making the warning you've reported here), but we've rolled out a bunch of updates that make that call to the agent happen from the SDK itself. Example of this https://github.com/percy/percy-puppeteer/pull/36

Upgrade to the latest @percy/puppeteer SDK and you should be all set.

Thanks,
David.

toolmantim:
Thanks David!

PR #34, opened 2018-12-09
[WIP] Promisify XMLHttpRequest

djones:
Convert synchronous `XMLHttpRequest`s into async ones by promisifying existing calls. This implementation specifically avoids `async`/`await` syntax as this client code has to be highly compatible across various test browsers.

Resolves issue #30

### DIFF
--- a/src/percy-agent-client/percy-agent-client.ts
+++ b/src/percy-agent-client/percy-agent-client.ts
@@ -9,28 +9,46 @@ export class PercyAgentClient {
     this.healthCheck()
   }
 
-  post(path: string, data: any) {
+  post(path: string, data: any): Promise <any> {
     if (!this.agentConnected) {
       console.warn('percy agent not started.')
-      return
+      return Promise.resolve()
     }
 
-    this.xhr.open('post', `${this.agentHost}${path}`, false) // synchronous request
-    this.xhr.setRequestHeader('Content-Type', 'application/json')
-    this.xhr.send(JSON.stringify(data))
+    return this._makeRequest(
+      `${this.agentHost}${path}`,
+      'post',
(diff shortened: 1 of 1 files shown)


src/percy-agent-client/percy-agent-client.ts
```diff
@@ -9,28 +9,46 @@ export class PercyAgentClient {
...
-    try {
-      this.xhr.open('get', `${this.agentHost}/percy/healthcheck`, false)
-      this.xhr.onload = () => {
-        if (this.xhr.status === 200) {
-          this.agentConnected = true
+  healthCheck(): Promise <any> {
+    return this._makeRequest(
+      `${this.agentHost}/percy/healthcheck`, 'get')
+    .then(() => {
+      this.agentConnected = true
```
cadeParade: It looks like you handled an error coming back previously, with `this.agentConnected = false`. This is no longer present because there's no `catch`.

src/percy-agent-client/percy-agent-client.ts
```diff
@@ -9,28 +9,46 @@ export class PercyAgentClient {
     this.healthCheck()
   }
 
-  post(path: string, data: any) {
+  post(path: string, data: any): Promise <any> {
     if (!this.agentConnected) {
       console.warn('percy agent not started.')
-      return
+      return Promise.resolve()
```
cadeParade: Would you want this to return as a failure? As-is it would return as a successfully resolved promise, but you could also return `Promise.reject()` which would signal to upstream callers that the thing was not successful.

djones:
Closing this and might re-visit later... it's not as important anymore because we've changed where the communication happens with the agent to fix CSP issues.

PR #37, opened 2018-12-10
Enable JS in asset discovery when the snapshot itself has JS enabled

djones:
I've added some test coverage too for when snapshot calls are made with options.

This PR also renames `enableJavascript` to `enableJavaScript` but supports both. Eventually we should deprecate `enableJavascript`.

### DIFF
--- a/src/percy-agent-client/percy-agent.ts
+++ b/src/percy-agent-client/percy-agent.ts
@@ -32,8 +32,10 @@ export default class PercyAgent {
     this.client.post('/percy/snapshot', {
       name,
       url: documentObject.URL,
+      // enableJavascript is deprecated. Use enableJavaScript
       enableJavaScript: options.enableJavaScript || options.enableJavascript,
       widths: options.widths,
+      // minimumHeight is deprecated. Use minHeight
       minHeight: options.minHeight || options.minimumHeight,
       clientInfo: this.clientInfo,
       environmentInfo: this.environmentInfo,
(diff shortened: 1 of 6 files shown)


src/percy-agent-client/percy-agent.ts
```diff
@@ -32,7 +32,7 @@ export default class PercyAgent {
     this.client.post('/percy/snapshot', {
       name,
       url: documentObject.URL,
-      enableJavascript: options.enableJavascript,
+      enableJavaScript: options.enableJavaScript || options.enableJavascript,
```
djones: Support both of these but `enableJavascript` will be deprecated eventually.

src/services/asset-discovery-service.ts
```diff
@@ -35,11 +35,10 @@ export default class AssetDiscoveryService extends PercyClientService {
     profile('-> assetDiscoveryService.browser.newPage')
     this.page = await this.browser.newPage()
     await this.page.setRequestInterception(true)
-    await this.page.setJavaScriptEnabled(false)
     profile('-> assetDiscoveryService.browser.newPage')
   }
 
-  async discoverResources(rootResourceUrl: string, domSnapshot: string): Promise<any[]> {
+  async discoverResources(rootResourceUrl: string, domSnapshot: string, enableJavaScript = false): Promise<any[]> {
```
timhaines: I dunno TS, but reading this makes me wonder if `enableJavaScript` is missing a type declaration?

djones: @timhaines yeah on the surface that does stand out. TypeScript just infers the type from the default value so you don't have to declare the type here.

Here's what my editor shows me:

screen shot 2018-12-10 at 2 18 06 pm

PR #40, opened 2018-12-11
Add utility to retrieve filename of percy-agent.js file

anaulin:


src/index.ts
```diff
@@ -1,2 +1,3 @@
 import PercyAgent from './percy-agent-client/percy-agent'
+export * from './utils/agentjs-filename'
```
anaulin: I'm not 100% sure this is the best way to export this from here, so would appreciate and extra set of checks here, @djones 🙏🏼

anaulin: Yeah, that's what I assumed.

Good idea on testing it with an SDK. Let me do that and come back to you here.

anaulin:
@djones this is what an updated SDK looks like: https://github.com/percy/percy-puppeteer/pull/28

As intended, we can do:
```typescript
import { agentJsFilename } from '@percy/agent'
```

and then call `agentJsFilename()`

PR #52, opened 2019-01-15
Add a mode in which PercyAgent does not post the domSnapshot to the service itself

anaulin:


src/percy-agent-client/client-options.ts
```diff
@@ -1,6 +1,7 @@
 export interface ClientOptions {
   clientInfo?: string,
   environmentInfo?: string,
+  postSnapshotDirectly?: boolean,
```
anaulin: This is not the best name, in terms of clarity, but I haven't come up with anything better yet.

src/percy-agent-client/percy-agent.ts
```diff
@@ -6,41 +6,52 @@ export default class PercyAgent {
   clientInfo: string | null
   environmentInfo: string | null
   xhr: any
+  postSnapshotDirectly: boolean
   port: number
   domTransformation: any | null
-  client: PercyAgentClient
+  client: PercyAgentClient | null
```
anaulin: This is something for discussion: If we don't want `PercyAgent` to be posting directly, then it doesn't need the `xhr` and it doesn't need the `client`. Do we still create and set those fields, or not?

Right now I've chosen to go with the path where, if we're not going to be posting directly, we leave those null. An advantage of this is that now you can instantiate `PercyAgent` from an environment where there isn't an `XMLHTTPRequest` available. The disadvantage is that now we have to check if those fields are null, and it is not the most elegant thing.

Robdel12:
Hm, the more I think about this the more I think this should be the default method. 🤔

anaulin:
@Robdel12 yeah, i agree so much that in this PR it is in fact the default method -- i.e. we will default to `postSnapshotDirectly = true` unless the constructor call explicitly tells us not to.

anaulin:
Actually, after talking to @djones , seems like we don't want to make any changes that would default to a different behavior. So the default behavior in the agent code should still be to post up the dom snapshot directly.

timhaines:
I'd like to understand that last comment a little better.  Is it trade-off between "not changing things for existing users" vs "having it work more reliably for the out-of-the-box setup going forward"?

src/percy-agent-client/client-options.ts
djones: * `handleSnapshotPosting`
* `handleAgentCommunication`

src/percy-agent-client/percy-agent.ts
djones: I think that's a good tradeoff. Seems reasonable to me to avoid creating objects that won't be used.

src/percy-agent-client/percy-agent.ts
```diff
@@ -6,41 +6,53 @@ export default class PercyAgent {
...
-      this.xhr,
-    )
+    if (this.postSnapshotDirectly) {
+      this.xhr = options.xhr || XMLHttpRequest
+      this.client = new PercyAgentClient(
+        `http://localhost:${this.port}`,
+        this.xhr,
+      )
+    } else {
+      this.xhr = null
```
djones: Do we need to actually set these as null? I would assume that's their default value.

If not that can be done inline with the definition. E.g.

```
client: PercyAgentClient | null = null
```

src/percy-agent-client/client-options.ts
anaulin: Since we're going to be doing the healthchecking ourselves too, we can go with `handleAgentCommunication`, I guess.

djones:
@timhaines @Robdel12 

> Is it trade-off between "not changing things for existing users" vs "having it work more reliably for the out-of-the-box setup going forward"?

Nope.

This new mode should be disabled by default because the `@percy/agent` changes are only one half of the puzzle. In order for this to work, each SDK must undergo changes like this https://github.com/percy/percy-cypress/pull/32

So we'll very purposely opt-in each SDK to use this new mode as it gains support for it.

Turning this mode on by default in `@percy/agent` alone doesn't fix CSP issues for all SDKs. If we wanted to have it on by default, we'd have to go back an update all the existing `@percy/agent` SDKs to turn it off right in the constructor until support is provided at SDK level.

anaulin:
@djones I addressed all your review comments, and went with the `handleAgentCommunication` name.

PR #54, opened 2019-01-16
Reject resource hint requests

### LINKED ISSUE #53: Resource hints cause rendering issues
Resource hints, https://css-tricks.com/prefetching-preloading-prebrowsing/, cause rendering issues as seen by this build https://percy.io/Attest-test/maker-fe/builds/1327395

The reason for this is because the resource hints on the first test are fetched, but this does not guarantee the content is fetched. It is just telling the browser to do something with the request optimising modern browser performance. 

For example if loading this page and taking a print screen will render fine because all the required assets are available as a full request with preload. However the second example will not render fine as the stylesheet has been cached from the previous page in the prefetch hook, however this request does not return any content as it is a resource hint and not required for the page to render. 

```HTML
<head>
    <link rel="prefetch" href="/assets/css/unique.css">
    <link rel="prefetch" href="/assets/css/secondary.css">
    <link rel="preload" href="/assets/css/shared.css" as="css">
    <link rel="stylesheet" href="/assets/css/shared.css">
</head>
```

```HTML
<head>
  <link rel="stylesheet" href="/assets/css/unique.css">
</head>
```

Essentially resource hints are not required to be loaded, and for the purpose of percy these hints can be ignored as they will not affect the rendering of the page and in the example of `<link rel="prefetch" href="/assets/css/secondary.css">` there is no use so percy has no need to story this asset.

blake-newman:
With resource hints in asset discovery, there is no guarantee that
the request will result in content being responded.

Add `shouldRequestResolve` method to determine if a request should
continue or be aborted in pupetteer.

If an asset is a resource hint then there is a purpose within the
request header, if this exists and is of a specific value we can
detirmine the request shouldn't resolve

Assets in resource hints are not necessary for the page to render
correctly they are there to give the browser better performance in
production applications/websites for the purpose of percy this assets my
never be requested thus storing unused asssets.

Fixes #53

### DIFF
--- a/src/services/asset-discovery-service.ts
+++ b/src/services/asset-discovery-service.ts
@@ -55,15 +55,20 @@ export default class AssetDiscoveryService extends PercyClientService {
     await this.page.setJavaScriptEnabled(enableJavaScript)
 
     this.page.on('request', async (request) => {
+      if (!this.shouldRequestResolve(request)) {
+        await request.abort()
+        return
+      }
+
       if (request.url() === rootResourceUrl) {
         await request.respond({
           body: domSnapshot,
           contentType: 'text/html',
           status: 200,
         })
-      } else {
-        await request.continue()
       }
+
+      await request.continue()
     })
 
     this.page.on('response', async (response) => {
@@ -98,6 +103,22 @@ export default class AssetDiscoveryService extends PercyClientService {
     profile('-> assetDiscoveryService.discoverResources', {resourcesDiscovered: resources.length})
 
     return resources
+  }
+
+  shouldRequestResolve(request: puppeteer.Request) {
+    const requestPurpose = request.headers().purpose
+
+    switch (requestPurpose) {
+      case 'prefetch':
+      case 'preload':
+      case 'dns-prefetch':
+      case 'prerender':
+      case 'preconnect':
+      case 'subresource':
+        return false
+      default:
+        return true
+    }
   }
 
   async teardown() {


src/services/asset-discovery-service.ts
```diff
@@ -55,15 +55,20 @@ export default class AssetDiscoveryService extends PercyClientService {
...
         await request.respond({
           body: domSnapshot,
           contentType: 'text/html',
           status: 200,
         })
-      } else {
-        await request.continue()
       }
+
+      await request.continue()
```
anaulin: @blake-newman @djones I'm upgraded to `@percy/agent@0.1.10`, which includes this change, and now I get this stack trace in all my tests:
```
(node:94052) UnhandledPromiseRejectionWarning: Error: Request is already handled!
    at assert (/Users/anaulin/src/github.com/percy/percy-puppeteer/node_modules/puppeteer/lib/helper.js:278:11)
    at Request.continue (/Users/anaulin/src/github.com/percy/percy-puppeteer/node_modules/puppeteer/lib/NetworkManager.js:398:5)
    at Request.<anonymous> (/Users/anaulin/src/github.com/percy/percy-puppeteer/node_modules/puppeteer/lib/helper.js:145:23)
    at Page.page.on (/Users/anaulin/src/github.com/percy/percy-puppeteer/node_modules/@percy/agent/dist/services/asset-discovery-service.js:51:35)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
(node:94052) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 18)
``` 

Looking that line 51 in the JS version is this line 71 in this file. Haven't investigated it in depth yet, but looks like a bug.

anaulin: Are we missing a `return` between lines 68 and 69?

djones: That seems very likely @anaulin.

anaulin: Fix: https://github.com/percy/percy-agent/pull/57/files

PR #75, opened 2019-02-05
Asset Discovery Skip Reason

djones:
Break out each reason we'd skip an asset from being discovered and output that to the log

### DIFF
--- a/src/services/response-service.ts
+++ b/src/services/response-service.ts
@@ -33,7 +33,7 @@ export default class ResponseService extends PercyClientService {
 
     if (request.url() === rootResourceUrl) {
       // Always skip the root resource
-      logger.debug(`Skipping [is_root_resource]: ${response.url()}`)
+      logger.debug(`Skipping [is_root_resource]: ${request.url()}`)
       return
     }
 
@@ -45,7 +45,7 @@ export default class ResponseService extends PercyClientService {
 
     if (!request.url().startsWith(rootUrl)) {
       // Disallow remote resource requests.
-      logger.debug(`Skipping [is_remote_resource]: ${response.url()}`)
+      logger.debug(`Skipping [is_remote_resource]: ${request.url()}`)
       return
     }
 
--- a/src/services/response-service.ts
+++ b/src/services/response-service.ts
@@ -31,15 +31,27 @@ export default class ResponseService extends PercyClientService {
     const parsedRootResourceUrl = new URL(rootResourceUrl)
     const rootUrl = `${parsedRootResourceUrl.protocol}//${parsedRootResourceUrl.host}`
 
-    if (
+    if (request.url() === rootResourceUrl) {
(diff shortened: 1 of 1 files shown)


src/services/response-service.ts
```diff
@@ -31,15 +31,27 @@ export default class ResponseService extends PercyClientService {
...
+    }
+
+    if (!this.ALLOWED_RESPONSE_STATUSES.includes(response.status())) {
       // Only allow 2XX responses:
-      || !this.ALLOWED_RESPONSE_STATUSES.includes(response.status())
-      || !request.url().startsWith(rootUrl) // Disallow remote resource requests.
-      || !response.url().startsWith(rootUrl) // Disallow remote redirects.
-      ) {
-      logger.debug(`Skipping: ${response.url()}`)
+      logger.debug(`Skipping [disallowed_response_status]: ${response.url()}`)
```
anaulin: In this case might be nice to log the response status we got.

Robdel12: I was going to say the same thing :D

djones: included with bf131e59a7898e6f3afbd4cae85196ed1e52b62e

src/services/response-service.ts
```diff
@@ -31,15 +31,27 @@ export default class ResponseService extends PercyClientService {
     const parsedRootResourceUrl = new URL(rootResourceUrl)
     const rootUrl = `${parsedRootResourceUrl.protocol}//${parsedRootResourceUrl.host}`
 
-    if (
+    if (request.url() === rootResourceUrl) {
       // Always skip the root resource
-      request.url() === rootResourceUrl
+      logger.debug(`Skipping [is_root_resource]: ${response.url()}`)
```
timhaines: Should this be request.url()?

src/services/response-service.ts
```diff
@@ -31,15 +31,27 @@ export default class ResponseService extends PercyClientService {
...
-      || !response.url().startsWith(rootUrl) // Disallow remote redirects.
-      ) {
-      logger.debug(`Skipping: ${response.url()}`)
+      logger.debug(`Skipping [disallowed_response_status_${response.status()}]: ${response.url()}`)
+      return
+    }
+
+    if (!request.url().startsWith(rootUrl)) {
+      // Disallow remote resource requests.
+      logger.debug(`Skipping [is_remote_resource]: ${response.url()}`)
```
timhaines: Should this be request.url() ?

src/services/response-service.ts
djones: Fixed with f0749262a8cc68828db9c0278aabf8ad1c1b8ee7

src/services/response-service.ts
djones: Fixed with f0749262a8cc68828db9c0278aabf8ad1c1b8ee7

PR #76, opened 2019-02-05
Serialize CSSOM into DOM

anaulin:
@djones @Robdel12 for review

### DIFF
--- a/src/percy-agent-client/percy-agent.ts
+++ b/src/percy-agent-client/percy-agent.ts
@@ -82,7 +82,8 @@ export default class PercyAgent {
     return `<!DOCTYPE ${doctype.name}` + publicDeclaration + systemDeclaration + '>'
   }
 
-  private serializeInputElements(domClone: HTMLElement): HTMLElement {
+  private serializeInputElements(doc: HTMLDocument): HTMLDocument {
+    const domClone = doc.documentElement
     const inputNodes = domClone.getElementsByTagName('input')
     const inputElements = Array.prototype.slice.call(inputNodes) as HTMLInputElement[]
 
@@ -99,14 +100,15 @@ export default class PercyAgent {
       }
     })
 
-    return domClone
+    return doc
   }
 
-  private stabilizeDOM(doc: Document): HTMLElement {
-    const docWithSerializedCSSOM = serializeCssOm(doc)
-    const stabilizedDOM = this.serializeInputElements(docWithSerializedCSSOM.documentElement)
+  private stabilizeDOM(doc: HTMLDocument): HTMLElement {
+    let stabilizedDOM = doc
+    stabilizedDOM = serializeCssOm(stabilizedDOM)
+    stabilizedDOM = this.serializeInputElements(stabilizedDOM)
     // more calls to come here
 
-    return stabilizedDOM
+    return stabilizedDOM.documentElement
   }
 }
--- a/src/percy-agent-client/percy-agent.ts
+++ b/src/percy-agent-client/percy-agent.ts
@@ -1,6 +1,7 @@
 import Constants from '../services/constants'
 import {ClientOptions} from './client-options'
 import {PercyAgentClient} from './percy-agent-client'
+import {serializeCssOm} from './serialize-cssom'
 import {SnapshotOptions} from './snapshot-options'
 
 export default class PercyAgent {
@@ -55,7 +56,7 @@ export default class PercyAgent {
 
   private domSnapshot(documentObject: Document): string {
     const doctype = this.getDoctype(documentObject)
-    const dom = this.stabilizeDOM(documentObject.documentElement as HTMLElement) as HTMLElement
+    const dom = this.stabilizeDOM(documentObject)
 
     let domClone = dom.cloneNode(true) as HTMLElement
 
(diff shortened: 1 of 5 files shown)


src/percy-agent-client/serialize-cssom.ts
```diff
@@ -0,0 +1,30 @@
...
+      const serializedStyles = [].slice
+        .call(styleSheet.cssRules)
+        .reduce((prev: string, cssRule: CSSRule) => {
+          return prev + cssRule.cssText
+        }, '')
+
+      // Append the serialized styles to the styleSheet's ownerNode to minimize
+      // the chances of messing up the cascade order.
+      const serializedSheet = document.createElement('style')
+      serializedSheet.type = 'text/css'
```
djones: it would be good to be able to identify this CSS as something _we_ injected in. Initially we'd talked about a data attribute or some type but I think [Percy specific CSS](https://docs.percy.io/docs/percy-specific-css) would be a particularly nice solution.

I guess it depends on what `serializedStyles` looks like. Could the serialized styles be simply wrapped in a `@media` block like this?

```css
@media only percy {
  /** CSSOM styles here **/
}
```

test/percy-agent-client/serialize-cssom.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  describe('dom snapshot', () => {
+    it('serializes JS-applied styles', () => {
+      const jsStyledDiv = document.getElementById('jsStyled') as HTMLInputElement
+      jsStyledDiv.style.background = 'red'
+
+      const domSnapshot = subject.snapshot('test snapshot')
+      const parsedDoc = (new DOMParser()).parseFromString(domSnapshot, 'text/html')
+
+      expect(parsedDoc.getElementById('jsStyled')!.style.background).to.contain('red')
```
djones: nice

src/percy-agent-client/serialize-cssom.ts
```diff
@@ -0,0 +1,30 @@
+// Take all the CSS created in the CSS Object Model (CSSOM), and inject it
+// into the DOM so Percy can render it safely in our browsers.
+// Design doc:
+// https://docs.google.com/document/d/1Rmm8osD-HwSHRpb8pQ_1wLU09XeaCV5AqMtQihk_BmM/edit
+export function serializeCssOm(document: HTMLDocument) {
```
djones: you could have this function return a `Document`, like this

```typescript
export function serializeCssOm(document: HTMLDocument): Document {
```

And then above in `percy-agent.ts`, change the call to:

```typescript
const docWithSerializedCSSOM = serializeCssOm(doc)
```

src/percy-agent-client/percy-agent.ts
```diff
@@ -101,8 +102,9 @@ export default class PercyAgent {
     return domClone
   }
 
-  private stabilizeDOM(dom: HTMLElement) {
-    const stabilizedDOM = this.serializeInputElements(dom)
+  private stabilizeDOM(doc: Document): HTMLElement {
+    const docWithSerializedCSSOM: Document = serializeCssOm(doc)
```
djones: I think about `stabilizeDOM()` in terms of a series of mutations on the input DOM. This call in particular feels a bit strange:

```typescript
this.serializeInputElements(docWithSerializedCSSOM.documentElement)
```

I'm wondering if something like this would be nicer:

```typescript
let stabilizedDOM = doc
stabilizedDOM = serializeCssOm(stabilizedDOM)
stabilizedDOM = this.serializeInputElements(stabilizedDOM)
```

And make a little update to `serializeInputElements` to take in a `HTMLDocument` not a document element.

Thoughts?

src/percy-agent-client/serialize-cssom.ts
anaulin: I just removed that type annotation in `percy-agent.ts`. (Had it there initially to make it clear that this is not just an `HTMLElement` the way `serializeInputElements` returns, but you're right that's it's probably unnecessary.)

src/percy-agent-client/percy-agent.ts
anaulin: Yeah, I agree that is nicer. Implemented and added to this PR.

anaulin:
@djones comments address, PTAL

src/percy-agent-client/serialize-cssom.ts
anaulin: If we can't get that `@media` thing to work, since I'm wrapping this style sheet in an element I'm creating myself, I could add a custom attribute to it, e.g. `data-percy-serialized` or something like that.

anaulin: Ok, added `data-percy-cssom-serialized` to the serialized styles!

src/percy-agent-client/percy-agent.ts
```diff
@@ -55,7 +56,7 @@ export default class PercyAgent {
 
   private domSnapshot(documentObject: Document): string {
     const doctype = this.getDoctype(documentObject)
-    const dom = this.stabilizeDOM(documentObject.documentElement as HTMLElement) as HTMLElement
+    const dom = this.stabilizeDOM(documentObject)
```
djones: ❤️

test/percy-agent-client/serialize-cssom.test.ts
```diff
@@ -0,0 +1,21 @@
+import {expect} from 'chai'
+import * as sinon from 'sinon'
+import PercyAgent from '../../src/percy-agent-client/percy-agent'
+import Constants from '../../src/services/constants'
+
+describe('serializeCssOm', () => {
+  const subject: PercyAgent = new PercyAgent({ handleAgentCommunication: false })
```
djones: ha, kinda nice that `handleAgentCommunication` made testing better here

Robdel12:
This is soooo gooodddd! 🔥 🏁

PR #79, opened 2019-02-07
Add integration test

anaulin:


test/integration/agent-integration.test.ts
```diff
@@ -0,0 +1,44 @@
...
+      args: [
+        '--disable-gpu',
+        '--single-process',
+      ],
+    })
+    page = await browser.newPage()
+  })
+
+  after(async () => {
+    browser!.close()
```
anaulin: @djones I don't love having to use `!` here and on lines 40 and 41, but I did not see a better way of getting Typescript to stop complaining about how `browser` and `page` could be `null`. Let me know if there is some Typescript magik that I'm missing.

test/integration/agent-integration.test.ts
```diff
@@ -0,0 +1,44 @@
+import { describe } from 'mocha'
+import * as puppeteer from 'puppeteer'
+import AgentService from '../../src/services/agent-service'
```
djones: `AgentService` and `Constants` I think are unused?

test/integration/agent-integration.test.ts
djones: I think if you remove the two ` | null = null` bits from `let browser` and `let page` above you can remove all the `!`s

anaulin: Yes! The key for me is that I didn't realize that I can just skip assigning `null` in the declaration. 🤦🏼‍♀️

test/integration/agent-integration.test.ts
anaulin: Oops. Removed.

anaulin:
@djones addressed your comments, please have another 👀

anaulin:
@djones I added the extra test cases we talked about:
* more complex live sites
* a facility to serve and snapshot all the `.html` files in a `/testcases` directory
* test cases for CSSOM and input element serialization

Please have another look.

package.json
```diff
@@ -112,8 +114,9 @@
     "posttest": "npm run lint",
     "prepublishOnly": "npm run build && oclif-dev manifest && oclif-dev readme",
     "preversion": "npm run clean",
-    "test": "npm run build-client && PERCY_TOKEN=abc mocha --forbid-only \"test/**/*.test.ts\" --exclude \"test/percy-agent-client/**/*.test.ts\"",
+    "test": "npm run build-client && PERCY_TOKEN=abc mocha --forbid-only \"test/**/*.test.ts\" --exclude \"test/percy-agent-client/**/*.test.ts\" --exclude \"test/integration/**/*\"",
     "test-client": "mkdir -p dist-test/ && npm run build-client-test && testem ci --file ./test/percy-agent-client/testem.js",
+    "test-integration": "npm run build-client && node ./bin/run exec -- mocha test/integration/**/*.test.ts",
```
djones: `node ./bin/run exec` pretty neat :+1:

PR #80, opened 2019-02-14
Use SnapshotOptions for passing around things like enableJavaScript

anaulin:
This PR introduces no functional changes, and is only a small refactor on how we pass around things like `widths` and `enableJavaScript`.

I think this variant helps keep things a little more consistent throughout the codebase, and we'll need it in any case as part of the responsive asset work, which requires `discoverResources(...)` to be aware of the requested snapshot `widths`.

### DIFF
--- a/src/services/agent-service.ts
+++ b/src/services/agent-service.ts
@@ -2,6 +2,7 @@ import * as bodyParser from 'body-parser'
 import * as cors from 'cors'
 import * as express from 'express'
 import {Server} from 'http'
+import { SnapshotOptions } from '../percy-agent-client/snapshot-options'
 import configuration, {SnapshotConfiguration} from '../utils/configuration'
 import logger, {profile} from '../utils/logger'
 import {AgentOptions} from './agent-options'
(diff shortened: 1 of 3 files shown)


src/services/snapshot-service.ts
```diff
@@ -52,14 +53,12 @@ export default class SnapshotService extends PercyClientService {
...
-    enableJavaScript = false,
-    widths: number[] | null = null,
-    minHeight: number | null = null,
+    options: SnapshotOptions = {},
     clientInfo: string | null = null,
     environmentInfo: string | null = null,
   ): Promise<any> {
     const snapshotCreationPromise: Promise<any> = this.percyClient.createSnapshot(
-      this.buildId, resources, {name, widths, enableJavaScript, minimumHeight: minHeight, clientInfo, environmentInfo},
+      this.buildId, resources, { name, ...options, minimumHeight: options.minHeight, clientInfo, environmentInfo },
```
anaulin: Heads up that in this implementation we are passing to the client both `minimumHeight` and `minHeight` in the options object. I don't think it matters, though, and this version is cleaner than manually passing through each option just so that we can rename `minHeight` to `minimumHeight`.

src/services/agent-service.ts
```diff
@@ -68,20 +69,23 @@ export default class AgentService {
 
     if (!this.snapshotService) { return response.json({success: false}) }
 
+    const snapshotConfiguration = (configuration().snapshot || {}) as SnapshotConfiguration
+    const snapshotOptions: SnapshotOptions = {
+      widths: request.body.widths || snapshotConfiguration.widths,
+      enableJavaScript: request.body.enableJavaScript,
```
anaulin: @djones I notice that `SnapshotConfiguration` does _not_ have an `enableJavaScript` field. Should it?

djones: It's ok to add it if it makes things easier. Generally speaking we don't want to encourage people to `enableJavaScript` though so that's why it was left off. But I don't feel super strongly about that.

anaulin: If we prefer people not to use it, then let's leave it off until a time when we actually do need it.

PR #81, opened 2019-02-15
Discover assets for all snapshot widths in parallel

anaulin:
Do asset discovery for all desired widths, in parallel, using a pool of Puppeteer `page` objects (i.e. browser tabs). The pool is of fixed size, and sized so that we will have at least one `page` available for each desired width, so we don't need to keep track of which `page` is currently "in use".

This change adds a little bit of latency to snapshots that request more than one width, in exchange for discovering responsive assets.


anaulin:
This change is also creating a visual diff that needs approving: https://percy.io/percy/percy-agent/builds/1452388?utm_campaign=percy&utm_content=percy-agent&utm_source=github_status

PR #85, opened 2019-02-21
Append CSSOM to owner node

Robdel12:
## What is this?

When serializing the CSSOM we're currently creating a `<style>` tag and appending it to the `<style>` tag that owns the CSSOM styles. This change ensures the serialized CSS from the CSSOM is appended to the `<style>` tag which the styles originated from rather than nesting `<style>` tags.

You can see this causing an issue in this example repo: https://github.com/montezume/percy-puppeteer-emotion/pull/1

This build was ran on that repo with this change: https://percy.io/robdel12/the-garbage-collection/builds/1470912 

### Pictures

**Before**

image

**After**

image


anaulin:
Sorry about the bug, and thanks for the fix, @Robdel12 !  (Now I'm curious as to why the buggy version didn't work in some cases, but works in others.)

PR #86, opened 2019-02-21
Serialize textareas for snapshots

Robdel12:
## What is this?

Previously we were only serializing input elements. This adds textareas to that list. They're a bit of a special element. I couldn't get it to work locally when setting the `value`, `text`, or anything else besides `innerText`. Which seems fine. 

[I splunked through how jQuery does it and they're setting the `textContent`](https://github.com/jquery/jquery/blob/master/src/manipulation.js#L334). I created a little jsbin to illustrate the differences: https://jsbin.com/hiquvuzele/1/edit?html,js,output
I have no issues changing `innerText` to `textContent` if that's something we want to do.

### DIFF
--- a/src/percy-agent-client/percy-agent.ts
+++ b/src/percy-agent-client/percy-agent.ts
@@ -96,7 +96,8 @@ export default class PercyAgent {
         }
         break
       case 'textarea':
-        elem.setAttribute('text', elem.value)
+        // setting text or value does not work but innerText does
+        elem.innerText = elem.value
       default:
         elem.setAttribute('value', elem.value)
       }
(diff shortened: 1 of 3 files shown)


src/percy-agent-client/percy-agent.ts
```diff
@@ -84,17 +84,20 @@ export default class PercyAgent {
...
       switch (elem.type) {
       case 'checkbox':
       case 'radio':
         if (elem.checked) {
           elem.setAttribute('checked', '')
         }
         break
+      case 'textarea':
+        // setting text or value does not work but innerText does
+        elem.innerText = elem.value
```
Robdel12: Any of these will work:

- `elem.innerText = elem.value`
- `elem.value = elem.value`
- `elem.textContent = elem.value`

djones:
We have to be pretty careful about what we include in the client code because it could be running in any old test browser... I was curious about the support for `querySelectorAll` but it looks well supported https://caniuse.com/#feat=queryselector so :+1:

Issue #92, opened 2019-03-08
No data found for resource with given identifier (Possible race condition)


Robdel12:
## Problem

As of `@percy/agent` version `0.1.15` we introduced responsive width detection, which renders the DOM snapshot we take in the widths that are specified to the SDK _concurrently_. When trying to snapshot some sites, this can cause a race condition where this error is thrown:

```
Error: Protocol error (Network.getResponseBody): No data found for resource with given identifier
    at Promise (/Users/robertdeluca/Projects/Work/percy-agent/node_modules/puppeteer/lib/Connection.js:186:56)
    at new Promise (<anonymous>)
    at CDPSession.send (/Users/robertdeluca/Projects/Work/percy-agent/node_modules/puppeteer/lib/Connection.js:185:12)
    at _contentPromise._bodyLoadedPromise.then (/Users/robertdeluca/Projects/Work/percy-agent/node_modules/puppeteer/lib/NetworkManager.js:591:45)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

Googling that error returns a few different issues that might be helpful:

- https://github.com/GoogleChrome/puppeteer/issues/2258
- https://github.com/GoogleChrome/puppeteer/issues/1274
- https://github.com/GoogleChrome/puppeteer/issues/2176

You can recreate this issue repeatedly with this puppeteer script:

```js
const puppeteer = require('puppeteer');
const {percySnapshot} = require('@percy/puppeteer');

(async () => {
  const browser = await puppeteer.launch({ headless: true });
  const page = await browser.newPage();

  await page.goto('https://www.nash.io/company/careers');

  await percySnapshot(page, 'Does this snapshot?');
  await browser.close();
})();
```

If you downgrade `@percy/agent` to `v0.1.14` the above snippet captures the snapshot properly. Singling out a width also does not help in the latest version. 

### Where does this blow up in agent? 💥 

Given the above puppeteer script, this error comes up on an `.mp4` video file when calling `.buffer()`on the response: https://github.com/percy/percy-agent/blob/master/src/services/response-service.ts#L70

![image](https://user-images.githubusercontent.com/2072894/54061193-6d528c80-41c5-11e9-9746-ccaa2ba57b83.png)

anaulin:
Digging deeper into what the agent is doing in this case, here is a Puppeteer-only script that reproduces this issue:
```javascript
const puppeteer = require('puppeteer');

const URL = 'https://www.nash.io/company/careers';


(async() => {

[...]

PR #94, opened 2019-03-11
Fix missing 'await' on promise

anaulin:
I believe this is a fix for https://github.com/percy/percy-agent/issues/92.

I've added an integration test case against the nash.io URL that @Robdel12 provide in his original bug report, but no local test case -- I don't know yet how doable might it be to reproduce this with a local test case, because I think the issue is loading a resource that is very large and takes a long time to load (and the missing `await` makes it so that we hit the race condition in that case). Let me know if you have opinions either way on trying to add a "local" test case or not. (Or if you have any thoughts on how to test this better, to make sure it doesn't regress again.)


anaulin:
@djones note new visual diff on this PR, because of the new live site snapshot: https://percy.io/percy/percy-agent/builds/1549496?utm_campaign=percy&utm_content=percy-agent&utm_source=github_status

anaulin:
Actually, never mind, seems like this didn't fully fix the issue. 😿

anaulin:
Closing this PR, as it was a red herring. I have a pure Puppeteer reproduce example for this issue, which I've posted on an existing Puppeteer issue around this problem: https://github.com/GoogleChrome/puppeteer/issues/1274#issuecomment-471793396

PR #97, opened 2019-03-13
Implement 'scope' feature

anaulin:
[The implementation and basic tests are complete in this PR, but before merging I'd like to add a couple more test cases for the error paths of this new code. Consider this an "early preview" PR.]

The main decision that I had to make in this implementation is that I put the scoping logic in `percy-agent.ts` (i.e. in `percy-agent.js`), so the scoping will happen in the browser. The main downside of this is that SDKs that do their own posting of the snapshot to the agent (i.e. the ones that set `handleAgentCommunication: false`) might need to be modified to make sure that we are passing the `scope` option into the in-browser operation of getting the DOM snapshot. It doesn't seem like a huge change, but just something to be aware of as we upgrade the SDKs to use the new version of the agent with this `scope` functionality.

I did originally go down the path of implementing this within `snapshot-service`, but since that code does not run in a browser, it does not have things like `DOMParser`, which led me in turn to consider depending on https://www.npmjs.com/package/xmldom to do the parsing from a snapshot string back into a DOM. It turned out that the DOM you get from the xmldom package doesn't have e.g. `querySelectorAll`, and generally I wasn't super comfortable with the idea of simulating operations on a Document outside of a browser, and depending on third-party packages that provide that.

Another decision in this PR is that, if there are multiple matching selectors, I snapshot all of them. Not sure if that's the behavior we want from this feature or not. Let me know.

### DIFF
--- a/src/percy-agent-client/percy-agent.ts
+++ b/src/percy-agent-client/percy-agent.ts
@@ -1,8 +1,8 @@
 import Constants from '../services/constants'
-import {ClientOptions} from './client-options'
-import {PercyAgentClient} from './percy-agent-client'
-import {serializeCssOm} from './serialize-cssom'
-import {SnapshotOptions} from './snapshot-options'
+import { ClientOptions } from './client-options'
+import { PercyAgentClient } from './percy-agent-client'
+import { serializeCssOm } from './serialize-cssom'
+import { SnapshotOptions } from './snapshot-options'
 
 export default class PercyAgent {
   clientInfo: string | null
@@ -34,7 +34,7 @@ export default class PercyAgent {
 
   snapshot(name: string, options: SnapshotOptions = {}) {
     const documentObject = options.document || document
-    const domSnapshot = this.domSnapshot(documentObject)
+    const domSnapshot = this.domSnapshot(documentObject, options.scope)
 
     if (this.handleAgentCommunication && this.client) {
       this.client.post(Constants.SNAPSHOT_PATH, {
@@ -47,6 +47,7 @@ export default class PercyAgent {
         minHeight: options.minHeight || options.minimumHeight,
         clientInfo: this.clientInfo,
         environmentInfo: this.environmentInfo,
+        scope: options.scope,
         domSnapshot,
       })
     }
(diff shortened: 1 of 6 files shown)


anaulin:
Seems like there are some unexpected diffs because of some autoformatting. Sorry about that, I'll fix that when I'm back on Thursday.

src/percy-agent-client/percy-agent.ts
```diff
@@ -47,19 +47,24 @@ export default class PercyAgent {
...
 
-  private domSnapshot(documentObject: Document): string {
+  private domSnapshot(documentObject: Document, scope?: string): string {
     const doctype = this.getDoctype(documentObject)
     const dom = this.stabilizeDOM(documentObject)
 
     let domClone = dom.cloneNode(true) as HTMLElement
 
+    if (scope) {
+      domClone = this.scopeDomSnapshot(domClone, scope)
```
djones: nice clear function name here :+1:

src/percy-agent-client/percy-agent.ts
```diff
@@ -114,4 +127,29 @@ export default class PercyAgent {
...
+
+      while (body.firstChild) {
+        body.removeChild(body.firstChild)
+      }
+      // Append only the scoped elements back to the body.
+      elements.forEach((el) => body!.appendChild(el))
+
+      return domClone
+    } catch (error) {
+      console.log(`[percy] Could not scope snapshot to: ${scope} ; Error: ${error}`)
```
djones: Nice graceful handling here. I think that makes sense. If a bad selector is sent through, falling back to no scope behavior makes sense to me

src/percy-agent-client/percy-agent.ts
```diff
@@ -114,4 +127,29 @@ export default class PercyAgent {
 
     return stabilizedDOM.documentElement
   }
+
+  private scopeDomSnapshot(domClone: HTMLElement, scope: string): HTMLElement {
+    try {
+      const elements = Array.from(domClone.querySelectorAll(scope))
+      if (elements.length === 0) {
+        return domClone
```
djones: :+1:

Robdel12:
The first thing that jumps out to me is how are we handling asset discovery here? The issue I ran into when I spiked `scoped` was missing assets since I also did the scoping _before_ discovery. It made the assumption that all assets are loaded in the `head` (including JS), which I don't think is safe.

anaulin:
After careful discussion, we've decided to not move forward with supporting this feature in the agent. We will instead provide guidance for customers on how to accomplish this with the existing tools we already have, which will also give users more flexibility on the specific behaviors that they want to see from such a feature.

PR #98, opened 2019-03-16
Process assets on 'requestfinished' instead of 'response'

anaulin:
Suggested by a conversation on the Puppeteer issue referenced from https://github.com/percy/percy-agent/issues/92, this change will make asset discovery a little bit more robust, by only trying process an asset once the request for it has succeeded in the browser.

### DIFF
--- a/src/services/asset-discovery-service.ts
+++ b/src/services/asset-discovery-service.ts
@@ -154,10 +154,10 @@ export default class AssetDiscoveryService extends PercyClientService {
     // We could also listen on 'response', but then we'd have to check if it was successful.
     page.on('requestfinished', async (request) => {
       const response = request.response()
-      if (response != null) {
+      if (response) {
         // Parallelize the work in processResponse as much as possible, but make sure to
         // wait for it to complete before returning from the asset discovery phase.
-        const promise = this.responseService.processResponse(rootResourceUrl, response!, width)
+        const promise = this.responseService.processResponse(rootResourceUrl, response, width)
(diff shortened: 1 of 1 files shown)


src/services/asset-discovery-service.ts
```diff
@@ -150,12 +150,24 @@ export default class AssetDiscoveryService extends PercyClientService {
...
-      // Parallelize the work in processResponse as much as possible, but make sure to
-      // wait for it to complete before returning from the asset discovery phase.
-      const promise = this.responseService.processResponse(rootResourceUrl, response, width)
-      promise.catch(logError)
-      maybeResourcePromises.push(promise)
+    // Listen on 'requestfinished', which tells us a request completed successfully.
+    // We could also listen on 'response', but then we'd have to check if it was successful.
+    page.on('requestfinished', async (request) => {
+      const response = request.response()
+      if (response != null) {
```
Robdel12: Curious, does this compile out to `!==` and we can use `!=` because it's TypeScript?

djones: I think just changing this to:

```typescript
if (response) {
```

is probably the simplest approach. Took a quick poke through https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html

anaulin: Yeah, not sure why I wrote it that way, I blame too much language-switching. Fixed now (and was able to remove the non-null coercion after fixing the null check).

PR #108, opened 2019-03-27
Add support for 10mb+ DOM snapshots

djones:
So far, we have a failing test case.

### DIFF
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -4,4 +4,4 @@
 --watch-extensions ts
 --recursive
 --reporter spec
---timeout 5000
+--timeout 20000


Robdel12:
Nice, should we add all the way up to what the API limits?

djones:
@Robdel12 yeah I think if we do set a limit, we could do so in the API also to gracefully handle this. Just feeling the problem out at the moment.

djones:
ah so we do have that 15mb limit on any resources... so this inherently includes the root resource too. Here's me trying to send a 23mb DOM to the API:

```
[percy] StatusCodeError 413 - "<html>\r\n<head><title>413 Request Entity Too Large</title></head>\r\n<body bgcolor=\"white\">\r\n<center><h1>413 Request Entity Too Large</h1></center>\r\n<hr><center>nginx/1.14.0</center>\r\n</body>\r\n</html>\r\n"
[percy] stopping percy...
[percy] waiting for 10 snapshots to complete...
[percy] StatusCodeError 400 - {"errors":[{"status":"bad_request","detail":"Finalizing snapshot 107578318 failed: cannot finalize before all snapshot resources are uploaded. This is likely a client library bug, please make sure that content for all SHAs in 'missing-resources' from the snapshot response are uploaded before calling finalize."}]}
[percy] done.
[percy] StatusCodeError 400 - {"errors":[{"status":"bad_request","detail":"Finalizing build 1642784 failed: cannot finalize before all snapshot resources are uploaded. This is likely a client error, please make sure that content for all SHAs in 'missing-resources' from the snapshot response are uploaded before calling finalize."}]}
[percy] finalized build #171: https://percy.io/percy/percy-agent/builds/1642784
```

djones:
Max resource size of 15mb is set here https://github.com/percy/percy-js/blob/3fff722d08329f6ecd6faeda7f1baf8b4799921a/src/utils.js#L7

PR #109, opened 2019-03-29
Carefully undo our serialized CSSOM from the customer DOM. Add tests.

anaulin:
This solution works in the general case, and this PR adds a test making sure that the behavior doesn't regress.

It is important to note, however, that in cases of malformed or otherwise degenerate underlying DOM, the clean up might not work perfectly, leaving behind pieces of serialized CSS.

I don't love having to do things this way -- it would be much cleaner to make a deep copy of the DOM and operate always on that. On the positive side, this implementation feels robust and understandable enough that it should be fairly maintainable.

Thoughts?

### DIFF
--- a/src/percy-agent-client/serialize-cssom.ts
+++ b/src/percy-agent-client/serialize-cssom.ts
@@ -32,8 +32,9 @@ export function serializeCssOm(document: HTMLDocument) {
 
 export function cleanSerializedCssOm(document: HTMLDocument) {
   // IMPORTANT: querySelectorAll(...) will not always work. In particular, in certain
-  // cases with malformed HTML (e.g. a <style> tag instead of another one), some of
-  // the elements we are looking for will not be returned.
+  // cases with malformed HTML (e.g. a <style> tag inside of another one), some of
+  // the elements we are looking for will not be returned. In that case, we will
+  // leave traces of ourselves in the underlying DOM.
   const nodes = document.querySelectorAll(`[${DATA_ATTRIBUTE}]`)
   Array.from(nodes).forEach((node: Element) => {
     node.removeAttribute(DATA_ATTRIBUTE)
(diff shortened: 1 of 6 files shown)


src/percy-agent-client/serialize-cssom.ts
```diff
@@ -15,14 +18,29 @@ export function serializeCssOm(document: HTMLDocument) {
...
       ownerNode.appendChild(document.createTextNode(serializedStyles))
     }
   })
 
   return document
 }
+
+export function cleanSerializedCssOm(document: HTMLDocument) {
+  // IMPORTANT: querySelectorAll(...) will not always work. In particular, in certain
+  // cases with malformed HTML (e.g. a <style> tag instead of another one), some of
```
Robdel12: `tag instead` should be `tag inside`?

test/percy-agent-client/percy-agent.test.ts
```diff
@@ -90,5 +90,18 @@ describe('PercyAgent', () => {
...
+      const originalHTML = document.documentElement.outerHTML
+
+      const firstDOMSnapshot = subject.snapshot('a snapshot')
+
+      const postSnapshotHTML = document.documentElement.outerHTML
+      expect(postSnapshotHTML).to.eq(originalHTML)
+      expect(postSnapshotHTML).to.not.contain('data-percy')
+
+      const secondDOMSnapshot = subject.snapshot('a second snapshot')
+      expect(secondDOMSnapshot).to.eq(firstDOMSnapshot)
```
djones: this is a super valuable test right here :+1:

anaulin:
Ah, @Robdel12 good point about the input elements, hadn't thought about that. Let me think about that in a follow-up?

anaulin:
I added a change to this PR that makes the tests a little cleaner, I think, @djones @Robdel12 do you want to have a final look?

PR #110, opened 2019-03-29
Allow assets from additional domains to be processed

green-arrow:
I recently messaged Percy.io via Intercom about an issue we are having with images / logos not appearing in our screenshots.

Our assets are served up from a separate domain, which percy is set up to ignore to reduce the overall number of assets that have to be processed. Unfortunately for us, our testing occurs behind a corporate VPN so those assets are not available to percy.

This PR introduces a new `-a` / `--asset-domains` flag, which takes in a comma separated list of asset domains that should be processed in addition to the root domain.

Please let me know if this is something you are interested in or if you need any additional information / work from me.

Thanks for making an awesome service!

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -29,6 +29,11 @@ export default class Start extends PercyCommand {
       default: Constants.PORT,
       description: 'port',
     }),
+    'asset-domains': flags.string({
+      char: 'a',
+      default: '',
+      description: 'additional domains for asset discovery',
+    }),
   }
 
   async run() {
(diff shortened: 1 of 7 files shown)


green-arrow:
Bumping this PR 😄

fotinakis:
@green-arrow were you able to test this? I actually wouldn't expect this to work, based on how we internally support rendering domains. Going to think a bit more about this but would love to hear if this actually fixed the issue for you in practice.

green-arrow:
@fotinakis - I did not actually end up testing this. I didn't realize that there was more internal work done on the Percy side that needed this information. Feel free to close this 😄

PR #111, opened 2019-03-29
Clean up serialized inputs after snapshotting

anaulin:
I've also made the logic of the serialization a little bit more conservative, doing the serialization only when it seems to be necessary. The drawback of that is that it makes the implementation a little bit less obvious (and thus more likely to contain a bug).

Finally, this change also moves input serialization into its own file (and associated test), similar to the structure we have for the CSSOM serialization code.

(This PR is the follow-up I promised in https://github.com/percy/percy-agent/pull/109)

### DIFF
--- a/test/percy-agent-client/serialize-input.test.ts
+++ b/test/percy-agent-client/serialize-input.test.ts
@@ -1,0 +1,47 @@
+import {expect} from 'chai'
+import * as sinon from 'sinon'
+import PercyAgent from '../../src/percy-agent-client/percy-agent'
+import Constants from '../../src/services/constants'
+import { htmlWithoutSelector } from '../helpers/html-string'
+
+describe('serializeInputElements', () => {
+  const subject: PercyAgent = new PercyAgent({ handleAgentCommunication: false })
+
+  it('serializes text input elements', () => {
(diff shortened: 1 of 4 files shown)


test/percy-agent-client/serialize-input.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  it('cleans up after itself', () => {
+    const preSnapshotHTML = htmlWithoutSelector(document, '#mocha')
+
+    subject.snapshot('test snapshot')
+
+    const postSnapshotHTML = htmlWithoutSelector(document, '#mocha')
+
+    expect(postSnapshotHTML).to.eq(preSnapshotHTML)
+    expect(postSnapshotHTML).to.not.contain('data-percy-input-serialized')
```
djones: Super useful tests here. One question, did you make sure this fails if we don't call `cleanSerializedInputElements`?

anaulin: Oh yeah. 👍 

Without the call to `cleanSerializedInputElements`:
Screen Shot 2019-03-29 at 4 51 50 PM

djones: Sweet, ship it!

PR #117, opened 2019-04-02
Make `percy-agent.js` browser & bundler ready

Robdel12:
## What is this?

When various SDKs use a bundler the entire percy agent `dist` was being bundled. This is bad because most of agent is a node process, and not meant for a browser. This [raised issues like this one](https://github.com/percy/percy-cypress/issues/58), where `fs` failed to be found (which is purely node). While debugging through this issue I also discovered we're not quite transpiling `percy-agent.js` (which is injected into the browser), so bundlers like webpack want attempt to bundle it when the test suite consuming the SDK runs. This is no good and goes 💥 . To fix this, there's many layers:

- Change the `main` field in `package.json` to point ONLY to `percy-agent.js`, which is injected into the browser. Bundlers use this field as a hint for what the packages entry is. _Not_ node. 
- Use rollup.js & babel to properly transpile the library for browser use. This is so the file is ready to go for any browser. 

## Future work

This introduces another bundler (rollup) to the equation. Currently we're using the typescript compiler and browserify.  I'd like to consolidate it all down to rollup in the future, but that's a bit out of scope of this PR and requires more time than I currently have. I think it would also simplify our test setup a TON.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,6 @@
       "delay": 1000
     }
   },
-  "browser": {
-    "winston": false
-  },
   "bugs": "https://github.com/percy/percy-agent/issues",
   "dependencies": {
     "@oclif/command": "^1",
@@ -106,7 +103,8 @@
     "agent"
   ],
   "license": "MIT",
-  "main": "dist/public/percy-agent.js",
+  "main": "dist/index.js",
+  "browser": "dist/public/percy-agent.js",
   "oclif": {
     "commands": "./dist/commands",
     "bin": "percy",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -94,7 +106,7 @@
     "agent"
   ],
   "license": "MIT",
-  "main": "dist/index.js",
+  "main": "dist/public/percy-agent.js",
```
Robdel12: This should be `browser` and leave `main` in place for other SDKs like `@percy/puppeteer`

Like: https://github.com/percy/percy-puppeteer/blob/master/lib/index.ts#L2

Robdel12:
I'm going to close this and reopen with the new scope. #119 takes care of some of the issues we were seeing but now we'd like to replace browserify with rollup

PR #118, opened 2019-04-02
Support post requests to the agent that are over 10mb

djones:
Set maxContentLength to 100mb to allow DOM snapshots over 10mb (the current default value).

Related: https://github.com/percy/percy-puppeteer/issues/70

### DIFF
--- a/src/utils/sdk-utils.ts
+++ b/src/utils/sdk-utils.ts
@@ -23,18 +23,20 @@ export async function isAgentRunning() {
   }
 
 export async function postSnapshot(body: any) {
-    const url = `http://localhost:${Constants.PORT}${Constants.SNAPSHOT_PATH}`
+    const URL = `http://localhost:${Constants.PORT}${Constants.SNAPSHOT_PATH}`
+    const ONE_HUNDRED_MB_IN_BYTES = 100_000_000
+
     return Axios({
       method: 'post',
-      url,
+      maxContentLength: ONE_HUNDRED_MB_IN_BYTES,
+      url: URL,
       data: body,
     } as any).then(() => {
       return true
-    }).catch((error) => {
+    }).catch(() => {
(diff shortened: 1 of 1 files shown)


src/utils/sdk-utils.ts
```diff
@@ -23,18 +23,20 @@ export async function isAgentRunning() {
...
+      url: URL,
       data: body,
     } as any).then(() => {
       return true
-    }).catch((error) => {
+    }).catch(() => {
       // This code runs in the context of the SDK, so the SDK needs to have LOG_LEVEL=debug
       // enabled for these logs to appear.
-      logger.debug(`Error posting snapshot to ${url} with body: ${body}`)
-      logger.debug(error)
```
djones: Specifically removing this line and the `${body}` above because of the monstrous log output you'll get when testing with larger DOMs.

src/utils/sdk-utils.ts
```diff
@@ -23,18 +23,20 @@ export async function isAgentRunning() {
...
+      logger.debug(`Error posting snapshot to ${URL}`)
```
Robdel12: Would it be worth having this as a log that isn't hidden behind `LOG_LEVEL=debug`?

djones: Yup. Will update.

djones: I changed it to use `logError`, which is a central way we have been rendering errors. It's defined here https://github.com/percy/percy-agent/blob/d01479d4387e5a9350199d163e0f387a835dab87/src/utils/logger.ts#L29-L32

It explicitly logs an error and then puts the full dump in debug. It outputs in the log like this (without `LOG_LEVEL=debug` being set)

```
    with large dom
[percy] Error Request body larger than maxBodyLength limit
[percy] Error posting snapshot to agent.
      ✓ snapshots without error (2002ms)
```

djones:
this shipped in @percy/agent 0.1.18

PR #120, opened 2019-04-03
feat: Add hidden snapshot command for future implementation

maprules1000:
## Description
This PR adds a hidden snapshot command, and a stubbed static snapshot service that will be fleshed out in a future PR.

This also includes tests for the snapshot command.

### DIFF
--- a/src/commands/snapshot.ts
+++ b/src/commands/snapshot.ts
@@ -17,7 +17,6 @@ export default class Snapshot extends PercyCommand {
   static examples = [
     '$ percy snapshot _mySite/',
     '$ percy snapshot _mySite/ --baseUrl "blog/"',
-    '$ percy snapshot _mySite/ --widths "320,780,1280"',
     '$ percy snapshot _mySite/ --ignore-folders "Tmp,_secrets,node_modules"',
   ]
 
@@ -31,11 +30,6 @@ export default class Snapshot extends PercyCommand {
       char: 'i',
       description: 'Comma-seperated string of folders to ignore. Ex: Tmp,_secrets,node_modules',
       default: '',
-    }),
-    'widths': flags.string({
-      char: 'w',
-      description: 'Comma-separated string of rendering widths for snapshots.',
-      default: '1280',
     }),
     'baseUrl': flags.string({
       char: 'b',
@@ -65,7 +59,6 @@ export default class Snapshot extends PercyCommand {
     const port = flags.port as number
     const portPlusOne = port + 1
     const networkIdleTimeout = flags['network-idle-timeout'] as number
-    const rawWidths = flags.widths as string
     const baseUrl = flags.baseUrl as string
     const rawIgnoreFolders = flags['ignore-folders'] as string
     const snapshotCaptureRegex = flags['snapshot-capture-regex'] as string
@@ -73,7 +66,6 @@ export default class Snapshot extends PercyCommand {
     // exit gracefully if percy will not run
     if (!this.percyWillRun()) { this.exit(0) }
 
-    const widths = rawWidths.split(',').map(Number)
     const ignoreFolders = rawIgnoreFolders ? rawIgnoreFolders.split(',') : undefined
 
     // start the agent service
@@ -83,7 +75,6 @@ export default class Snapshot extends PercyCommand {
     const options: StaticSnapshotOptions = {
       port: portPlusOne,
       staticAssetDirectory,
-      widths,
       baseUrl,
       snapshotCaptureRegex,
       ignoreFolders,
--- a/src/commands/snapshot.ts
+++ b/src/commands/snapshot.ts
@@ -16,7 +16,7 @@ export default class Snapshot extends PercyCommand {
 
   static examples = [
     '$ percy snapshot _site/',
-    '$ percy snapshot _site/ --baseUrl "blog/"',
+    '$ percy snapshot _site/ --base-url "blog/"',
     '$ percy snapshot _site/ --ignore-files "\.(blog|docs)$"',
   ]
 
@@ -28,7 +28,7 @@ export default class Snapshot extends PercyCommand {
     }),
     'ignore-files': flags.string({
       char: 'i',
-      description: 'Comma-seperated string of folders to ignore. Ex: Tmp,_secrets,node_modules',
+      description: 'Regular expression for matching the files to ignore.',
       default: '',
     }),
     'base-url': flags.string({
@@ -55,7 +55,7 @@ export default class Snapshot extends PercyCommand {
 
     const {args, flags} = this.parse(Snapshot)
 
-    const staticAssetDirectory = args.staticAssets as string
+    const snapshotDirectory = args.snapshotDirectory as string
     const port = flags.port as number
     const staticServerPort = port + 1
     const networkIdleTimeout = flags['network-idle-timeout'] as number
@@ -72,7 +72,7 @@ export default class Snapshot extends PercyCommand {
 
     const options: StaticSnapshotOptions = {
       port: staticServerPort,
-      staticAssetDirectory,
+      snapshotDirectory,
       baseUrl,
       snapshotFilesRegex,
       ignoreFilesRegex,
--- a/src/commands/snapshot.ts
+++ b/src/commands/snapshot.ts
@@ -2,6 +2,7 @@ import {flags} from '@oclif/command'
 import Constants from '../services/constants'
 import {StaticSnapshotOptions} from '../services/static-snapshot-options'
 import StaticSnapshotService from '../services/static-snapshot-service'
+import logger from '../utils/logger'
 import PercyCommand from './percy-command'
 
 export default class Snapshot extends PercyCommand {
@@ -16,7 +17,7 @@ export default class Snapshot extends PercyCommand {
 
   static examples = [
     '$ percy snapshot _site/',
-    '$ percy snapshot _site/ --base-url "blog/"',
+    '$ percy snapshot _site/ --base-url "/blog/"',
     '$ percy snapshot _site/ --ignore-files "\.(blog|docs)$"',
   ]
 
@@ -33,8 +34,8 @@ export default class Snapshot extends PercyCommand {
     }),
     'base-url': flags.string({
       char: 'b',
-      description: 'The path that the site will be deployed to on a production server. \
-      Use this if your site will be hosted at a non-root url.',
+      description: 'If your static files will be hosted in a subdirectory, instead \n' +
+      'of the webserver\'s root path, set that subdirectory with this flag.',
       default: '/',
     }),
     // from exec command. needed to start the agent service.
@@ -55,6 +56,8 @@ export default class Snapshot extends PercyCommand {
 
     const {args, flags} = this.parse(Snapshot)
 
+    const isWindows = process.platform === 'win32'
+
     const snapshotDirectory = args.snapshotDirectory as string
     const port = flags.port as number
     const staticServerPort = port + 1
@@ -65,6 +68,13 @@ export default class Snapshot extends PercyCommand {
 
     // exit gracefully if percy will not run
     if (!this.percyWillRun()) { this.exit(0) }
+
+    // check that base url starts with a slash and  exit if it is missing
+    if (isWindows) {
+      this.checkForWrappingSlashes(baseUrl, '\\')
+    } else {
+      this.checkForWrappingSlashes(baseUrl, '/')
+    }
 
     // start the agent service
     await this.agentService.start({port, networkIdleTimeout})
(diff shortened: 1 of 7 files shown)


test/commands/snapshot.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      chai.expect(staticSnapshotServiceStub.snapshotAll).to.have.callCount(1)
+      chai.expect(stdout).to.match(/\[percy\] percy has started./)
+    })
+
+    // todo: how to test that the flags are being passed to the service correctly?
+    xit('starts the snapshot service on the correct port')
+    xit('passes the correct args to the snapshotAll command')
+
+    // old way of testing flags
+    // it('passes the correct args to the static snapshot service', async () => {
```
maprules1000: I'm keeping this around as a point of reference of what is not being tested right now. Hopefully, after staring at it long enough, I will figure out how to test for these things.

maprules1000:
This PR is a stepping stone to #132 which will implement and test the underlying static snapshot service

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+    }),
+    'ignore-folders': flags.string({
+      char: 'i',
+      description: 'Comma-seperated string of folders to ignore. Ex: Tmp,_secrets,node_modules',
+      default: '',
+    }),
+    'widths': flags.string({
+      char: 'w',
+      description: 'Comma-separated string of rendering widths for snapshots. Ex: 320,1280',
+      default: '1280',
```
Robdel12: Should we rely on the defaults that are in the API? I remember seeing something that removed the defaults from agent in favor of that

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+      default: Constants.PORT,
+      description: 'port',
+    }),
+  }
+
+  async run() {
+    await super.run()
+
+    const {args} = this.parse(Snapshot)
+    const {flags} = this.parse(Snapshot)
```
Robdel12: I believe this could become one line:

```js
const {args, flags} = this.parse(Snapshot)
```

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+    'baseUrl': flags.string({
+      char: 'b',
+      description: 'The path that the site will be deployed to on a production server. \
+      Use this if your site will be hosted at a non-root url.',
+      default: '/',
+    }),
+    // from exec command. needed to start the agent service.
+    'network-idle-timeout': flags.integer({
+      char: 't',
+      default: 50,
```
Robdel12: Wonder if this should be moved to the constants?

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+      widths,
+      baseUrl,
+      snapshotCaptureRegex,
+      ignoreFolders,
+    }
+
+    const staticSnapshotService = new StaticSnapshotService(options)
+
+    // start the snapshot service
+    staticSnapshotService.start()
```
Robdel12: Missing an `await` here

src/services/static-snapshot-service.ts
```diff
@@ -0,0 +1,20 @@
+import logger from '../utils/logger'
+import {StaticSnapshotOptions} from './static-snapshot-options'
+
+export default class StaticSnapshotService {
```
Robdel12: As I read through this, I wonder if it would make sense to have the snapshot service extend the agent service? Or at least take care of the setup of the agent service in this constructor, so you don't have to do that in the `run` command method

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
+import {flags} from '@oclif/command'
+import Constants from '../services/constants'
+import {StaticSnapshotOptions} from '../services/static-snapshot-options'
+import StaticSnapshotService from '../services/static-snapshot-service'
+import PercyCommand from './percy-command'
+
+export default class Snapshot extends PercyCommand {
+  static description = 'Snapshot a directory of static site assets'
```
djones:
```suggestion
  static description = 'Snapshot a directory of webpages'
```

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+  static hidden = true
+
+  static args = [{
+    name: 'staticAssets',
```
djones:
```suggestion
    name: 'snapshotDirectory',
```

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+    description: 'The path to the compiled static asset directory.',
```
djones:
```suggestion
    description: 'A path to the directory you would like to snapshot',
```

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+    '$ percy snapshot _mySite/',
+    '$ percy snapshot _mySite/ --baseUrl "blog/"',
+    '$ percy snapshot _mySite/ --widths "320,780,1280"',
+    '$ percy snapshot _mySite/ --ignore-folders "Tmp,_secrets,node_modules"',
+  ]
+
+  static flags = {
+    'snapshot-capture-regex': flags.string({
+      char: 'c',
+      description: 'Regular expression for matching the files to snapshot. Defaults to: "\.(html|htm)$"',
```
djones: Running:

```
./bin/run snapshot --help
```

Shows you how this appears in help. Oclif will output the default values already so you don't need to include them in the description.

```suggestion
      description: 'Regular expression for matching the files to snapshot.',
```

src/commands/snapshot.ts
maprules1000: Yup, I agree. I was just thinking about that yesterday

src/commands/snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+      default: '\.(html|htm)$',
+    }),
+    'ignore-folders': flags.string({
+      char: 'i',
+      description: 'Comma-seperated string of folders to ignore. Ex: Tmp,_secrets,node_modules',
+      default: '',
+    }),
+    'widths': flags.string({
+      char: 'w',
+      description: 'Comma-separated string of rendering widths for snapshots. Ex: 320,1280',
```
djones: Running:

```
./bin/run snapshot --help
```

Shows you how this appears in help. Oclif will output the default values already so you don't need to include them in the description.

```suggestion
      description: 'Comma-separated string of rendering widths for snapshots.',
```

src/commands/snapshot.ts
djones: That would be great. @maprules1000 can you draw inspiration from the `Constants.PORT` usage below? Be sure to apply the same change to the other commands.

src/commands/snapshot.ts
djones: Nice catch :+1:

test/commands/snapshot.test.ts
maprules1000: Figured this out. The argument passing is tested in #137

src/services/static-snapshot-service.ts
maprules1000: Hmm, I feel like this conversation/decision might be better had on #137 since that fleshes out (and changes) the static snapshot service @Robdel12

src/commands/snapshot.ts
djones: yep, I think we should remove the `--widths` flag completely. You will be able to use the `.percy.yml` file to set them if you want to.

src/commands/snapshot.ts
```diff
@@ -0,0 +1,104 @@
...
+      char: 'i',
+      description: 'Comma-seperated string of folders to ignore. Ex: Tmp,_secrets,node_modules',
+      default: '',
+    }),
+    'widths': flags.string({
+      char: 'w',
+      description: 'Comma-separated string of rendering widths for snapshots.',
+      default: '1280',
+    }),
+    'baseUrl': flags.string({
```
djones: Suggestion to rename to `base-url` to keep consistent with Oclif's conventions.

src/commands/snapshot.ts
```diff
@@ -0,0 +1,104 @@
...
+  static hidden = true
+
+  static args = [{
+    name: 'snapshotDirectory',
+    description: 'A path to the directory you would like to snapshot',
+    required: true,
+  }]
+
+  static examples = [
+    '$ percy snapshot _mySite/',
```
djones: let's change the example dir to `_site/` (it happens to be the default for Jekyll)

src/services/constants.ts
```diff
@@ -1,5 +1,6 @@
 export default class Constants {
   static readonly PORT: number = 5338
+  static readonly NETWORK_IDLE_TIMEOUT: number = 50 // in milliseconds
```
djones: Fab

src/commands/snapshot.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  async run() {
+    await super.run()
+
+    const {args, flags} = this.parse(Snapshot)
+
+    const staticAssetDirectory = args.staticAssets as string
+    const port = flags.port as number
+    const portPlusOne = port + 1
```
djones:
```suggestion
    const staticServerPort = port + 1
```

src/commands/snapshot.ts
maprules1000: 👍

src/commands/snapshot.ts
maprules1000: 👍

src/commands/snapshot.ts
maprules1000: 👍

src/commands/snapshot.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  static args = [{
+    name: 'snapshotDirectory',
+    description: 'A path to the directory you would like to snapshot',
+    required: true,
+  }]
+
+  static examples = [
+    '$ percy snapshot _site/',
+    '$ percy snapshot _site/ --baseUrl "blog/"',
```
djones: The example here is now out of date:

```suggestion
    '$ percy snapshot _site/ --base-url "blog/"',
```

src/commands/snapshot.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  static flags = {
+    'snapshot-files': flags.string({
+      char: 'c',
+      description: 'Regular expression for matching the files to snapshot.',
+      default: '\.(html|htm)$',
+    }),
+    'ignore-files': flags.string({
+      char: 'i',
+      description: 'Comma-seperated string of folders to ignore. Ex: Tmp,_secrets,node_modules',
```
djones:
```suggestion
      description: 'Regular expression for matching the files to ignore from snapshotting.',
```

src/commands/snapshot.ts
```diff
@@ -0,0 +1,93 @@
...
+      description: 'port',
+    }),
+  }
+
+  async run() {
+    await super.run()
+
+    const {args, flags} = this.parse(Snapshot)
+
+    const staticAssetDirectory = args.staticAssets as string
```
djones:
```suggestion
    const staticAssetDirectory = args.snapshotDirectory as string
```

src/commands/snapshot.ts
```diff
@@ -0,0 +1,93 @@
...
+    // exit gracefully if percy will not run
+    if (!this.percyWillRun()) { this.exit(0) }
+
+    // start the agent service
+    await this.agentService.start({port, networkIdleTimeout})
+    this.logStart()
+
+    const options: StaticSnapshotOptions = {
+      port: staticServerPort,
+      staticAssetDirectory,
```
djones:
```suggestion
      snapshotDirectory,
```

src/services/static-snapshot-options.ts
```diff
@@ -0,0 +1,7 @@
+export interface StaticSnapshotOptions {
+    staticAssetDirectory: string,
```
djones:
```suggestion
    snapshotDirectory: string,
```

src/commands/snapshot.ts
```diff
@@ -0,0 +1,110 @@
+import {flags} from '@oclif/command'
+import Constants from '../services/constants'
+import {StaticSnapshotOptions} from '../services/static-snapshot-options'
+import StaticSnapshotService from '../services/static-snapshot-service'
+import logger from '../utils/logger'
+import PercyCommand from './percy-command'
+
+export default class Snapshot extends PercyCommand {
+  static description = 'Snapshot a directory of webpages'
```
timhaines: this feels weird to me.  Maybe 'Snapshot a directory of html files'?   Or 'Snapshot a folder containing a pre-built static website`?

maprules1000: I like the latter. Users can technically change the extension of files they want to snapshot

PR #121, opened 2019-04-03
chore: Add semantic-release for automatic package releases

djones:
This adds https://github.com/semantic-release/semantic-release support.

The default plugins that run include GitHub and NPM and I've configured CircleCI to have the correct tokens.

It's a wee bit hard to actually test this all works but the release step does seem to [run and bail out](https://circleci.com/gh/percy/percy-agent/1118?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) since it's not on master, which is to be expected.

From this point forward our pull request titles will need to [have certain keywords in them](https://semantic-release.gitbook.io/semantic-release/#commit-message-format).

PR #122, opened 2019-04-03
feat: Add `healthcheck` command

Robdel12:
## What is this?

In some instances it's helpful to have a CLI command that runs a check to make sure the Percy process is running. This will be especially helpful in the Cypress SDK where we're having struggles with providing a nice way to run a CLI command with the SDK:

- https://github.com/percy/percy-cypress/pull/67
- https://github.com/percy/percy-cypress/pull/69
- https://github.com/percy/percy-cypress/pull/72

## Usage

`$ percy health-check` or `$ percy health-check -p 5502`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       - checkout
       - restore_cache: &restore_cache
           keys:
-            - v1-yarn-{{checksum ".circleci/config.yml"}}-{{ checksum "yarn.lock"}}
+            - v1-yarn-{{checksum ".circleci/config.yml"}}-{{ checksum "package-lock.json"}}
             - v1-yarn-{{checksum ".circleci/config.yml"}}
       - run: .circleci/greenkeeper
       - run: yarn add -D nyc@11 @oclif/nyc-config@1 mocha-junit-reporter@1
--- a/src/commands/health-check.ts
+++ b/src/commands/health-check.ts
@@ -1,11 +1,10 @@
-import {flags} from '@oclif/command'
+import {Command, flags} from '@oclif/command'
 import * as colors from 'colors'
 import Constants from '../services/constants'
 import healthCheck from '../utils/health-checker'
-import PercyCommand from './percy-command'
 
-export default class HealthCheck extends PercyCommand {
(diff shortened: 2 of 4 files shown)


Robdel12:
I'll probably have to rebase this after #121 & maybe #120

src/commands/health-check.ts
```diff
@@ -0,0 +1,37 @@
...
+  static examples = [
+    '$ percy healthcheck',
+    '$ percy healthcheck --port 6884',
+  ]
+
+  async run() {
+    await super.run()
+
+    // If Percy is disabled or is missing a token, gracefully exit here
+    if (!this.percyWillRun()) { this.exit(0) }
```
djones: not super sure if this is actually relevant in this particular context. It's useful for gracefully disabling a command if Percy isn't going to run, but this is sort of different. `this.percyWillRun()` will also check for a `PERCY_TOKEN` but that's not relevant here.

src/commands/health-check.ts
```diff
@@ -0,0 +1,37 @@
+import {flags} from '@oclif/command'
+import * as colors from 'colors'
+import Constants from '../services/constants'
+import healthCheck from '../utils/health-checker'
+import PercyCommand from './percy-command'
+
+export default class HealthCheck extends PercyCommand {
```
djones: I think you might just be able to extend `Command` instead due to the nature of this command.

.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - checkout
       - restore_cache: &restore_cache
           keys:
-            - v1-yarn-{{checksum ".circleci/config.yml"}}-{{ checksum "yarn.lock"}}
+            - v1-yarn-{{checksum ".circleci/config.yml"}}-{{ checksum "package-lock.json"}}
```
djones: thank you

src/commands/health-check.ts
Robdel12: Yeah I debated if I should add it. It'd probably fail no matter what -- so I can remove it.

src/commands/health-check.ts
Robdel12: Really good idea -- there's a lot more in PercyCommand than I realized

PR #124, opened 2019-04-05
build: Introduce rollup to bundle `percy-agent.js` for browsers

Robdel12:
## What is this?

This PR introduces the rollup bundler so we can properly bundle the `percy-agent.js` file. Currently we transpile it with browserify but we don't bundle all of the code together into a single file. Since we're injecting this code into browsers it would be best to transpile it and then pull it together into a single file (so there are no `requires` or anything of that nature in the file).

This also introduces babel, which will soon be replacing all of browserify in this project. The community has essentially settled on babel as the ecosystem to transpile future JS code to JS that can run in the browser today.


.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+  "presets": [
+    "@babel/env",
+    "@babel/typescript"
+  ],
+  "plugins": [
+    "@babel/proposal-class-properties",
+    "@babel/proposal-object-rest-spread"
+]
```
djones: indentation fix

package.json
```diff
@@ -53,7 +55,9 @@
     "@types/mocha": "^5.2.5",
     "@types/nock": "^9.3.0",
     "@types/node": "^10.7.1",
+    "@types/sinon": "^5.0.1",
     "@types/sinon-chai": "^3.2.0",
+    "@semantic-release/git": "^7.0.8",
```
djones: thanks

package.json
```diff
@@ -109,7 +118,7 @@
   "repository": "percy/percy-agent",
   "scripts": {
     "build": "npm run clean && rm -rf dist && tsc && npm run build-client",
-    "build-client": "mkdir -p dist/public && browserify src/percy-agent-client/index.ts -p [ tsify --noImplicitAny ] --standalone PercyAgent > dist/public/percy-agent.js",
+    "build-client": "rollup -c",
```
djones: 😍

Robdel12:
Really curious what `build:` does with the release process! 👀

PR #136, opened 2019-04-11
build: Azure Pipelines

djones:
This PR moves us to Azure Pipelines which means we now get test coverage on Mac and Windows, in addition to the current Linux only test runs.

The test runs are configured to report test meta data including test coverage. Examples:

Screen Shot 2019-05-08 at 5 04 46 PM

Screen Shot 2019-05-08 at 5 04 53 PM


package.json
```diff
@@ -127,7 +127,8 @@
     "posttest": "npm run lint",
     "prepublishOnly": "npm run build && oclif-dev manifest && oclif-dev readme",
     "preversion": "npm run clean",
-    "test": "npm run build-client && PERCY_TOKEN=abc mocha --forbid-only \"test/**/*.test.ts\" --exclude \"test/percy-agent-client/**/*.test.ts\" --exclude \"test/integration/**/*\"",
+    "pretest": "npm run build-client",
```
Robdel12: 🔥 YES

src/services/process-service.ts
```diff
@@ -35,7 +36,11 @@ export default class ProcessService {
       const pid = this.getPid()
       fs.unlinkSync(ProcessService.PID_PATH)
 
-      process.kill(pid, 'SIGINT')
+      if (os.platform() === 'win32') {
+        childProcess.exec('taskkill /pid ' + pid + ' /T /F')
+      } else {
+        process.kill(pid, 'SIGHUP')
```
timhaines: You changed from using SIGINT to SIGHUP on non-windows?

djones: I think this might be a comment on an old commit 😄. Ended up splitting out `win32` as a separate step and non-windows on the existing `SIGHUP`.

djones:
Todos:

- [ ] Setup release step
- [ ] Remove CircleCI conf
- [ ] Remove AppVeyor conf

djones:
- [ ] add azure pipelines badge to readme

Robdel12:
Going to close this PR since it's pretty stale & we'll probably use GH actions instead (better wrapper over azure)

Issue #154, opened 2019-04-12
Dependency error on `@oclif/errors`


smacgaha:
When running `percy` from the runtime (not globally), an error is raised:

`Error: Cannot find module '@oclif/errors'`

I believe that this is because `@oclif/errors` is required in the binary, but is only installed as a devDependency.

No idea why this works when installed globally. I suspect it's related to the fact that `@oclif/errors` is also installed as a dependency of `@oclif/command`.

djones:
I think we can remove it completely since, as you point out, it is already a dependency of `@oclif/command`. Also I note that the latest Oclif example project [doesn't list it explicitly](https://github.com/oclif/example-multi-ts/blob/master/package.json#L10) as a dependency.

smacgaha:
Sure thing, I'll try that out

PR #155, opened 2019-04-12
fix: Remove @oclif/errors from devDependency list

smacgaha:
See https://github.com/percy/percy-agent/issues/154


djones:
@smacgaha as per discussion on #154 would you mind removing `@oclif/errors` completely? Then we should be good to merge.

smacgaha:
@djones I tested this locally and it worked fine; I think it's good to go.

djones:
Thanks @smacgaha! Thanks for the contribution 💜

PR #159, opened 2019-04-17
feat: Support static snapshots in `.percy.yml` configuration file

maprules1000:
## Description
This PR adds the static snapshot flags to the percy.yml file.

Since each flag has a default, the preference is to look for the yml file setting, and fall back to the flags passed in through the command line. Otherwise the defaults will always be used and the percy.yml settings will never be picked up.


src/utils/configuration.ts
```diff
@@ -22,7 +29,9 @@ const configuration = (relativePath = '.percy.yml'): Configuration => {
     // in a chain. snapshot specific options -> agent configuration -> default values
 
     const defaultConfiguration: Configuration = {
-      version: 1.0, snapshot: {},
+      version: 1.0,
```
maprules1000: Should this version be updated since it introduces `static_site`? It is still 100% compatible with the existing format

test/support/.percy.yml
```diff
@@ -2,3 +2,7 @@ version: 1
 snapshot:
   widths: [375, 1280]
   min-height: 1024
```
Robdel12: Pre-existing, but interesting this is not camel cased.

test/support/.percy.yml
```diff
@@ -2,3 +2,7 @@ version: 1
...
+static_site:
+  base-url: /blog
```
Robdel12: Wondering if all of these should be camel cased just like the other config options are:
- https://github.com/percy/percy-agent/blob/7e92365023f8869377ae4d9eb71f04c929941144/src/percy-agent-client/snapshot-options.ts
- https://github.com/percy/percy-agent/blob/7e92365023f8869377ae4d9eb71f04c929941144/src/services/agent-options.ts
- https://github.com/percy/percy-agent/blob/7e92365023f8869377ae4d9eb71f04c929941144/src/percy-agent-client/client-options.ts

src/utils/configuration.ts
Robdel12: I think so, 1.1?

src/utils/configuration.ts
```diff
@@ -7,9 +7,16 @@ export interface SnapshotConfiguration {
   'min-height'?: number,
 }
 
+export interface StaticSiteSnapshotConfiguration {
```
Robdel12: Wonder if this should be its own file to match the setup of the other config options

src/utils/configuration.ts
```diff
@@ -7,9 +7,16 @@ export interface SnapshotConfiguration {
...
+  'base-url'?: string,
+  'snapshot-files'?: string,
```
fotinakis: @djones maybe we name this `include-paths` and `exclude-paths`?

djones: These configuration names mirror the CLI flag names. I think it'd be good to keep those two consistent.

src/utils/configuration.ts
```diff
@@ -22,11 +29,12 @@ const configuration = (relativePath = '.percy.yml'): Configuration => {
     // in a chain. snapshot specific options -> agent configuration -> default values
 
     const defaultConfiguration: Configuration = {
-      version: 1.0, snapshot: {},
+      'version': 1.0,
```
djones: this should stay as `1.0`. These new options are completely compatible with the current configuration. We don't actually have any configuration version handling in place, this was just added for future proofing if we do find ourselves with breaking changes.

test/support/.percy.yml
djones: resolved with https://github.com/percy/percy-agent/pull/159/commits/2568731352e8dcdbd1095e19f163f9daec026ed5

PR #160, opened 2019-04-17
fix: Revert DOM clean up methods

Robdel12:
## Why? 

Given this JSBin: https://jsbin.com/?html,console,output

When we serialize the CSSOM into the `<style>` tag that the in-memory styles own and then _remove_ those styles, it triggers `deleteRule` on those styles in the CSSOM. Which means we're removing the text we injected AND all of the styles that live in memory. This breaks their app.

With the input cleaning, when we remove the value or checked attribute on the element, it's possible those elements lose their values entirely. For example, when we remove the `checked` attribute from checkboxes in the customers DOM, this will make the checkbox _no longer checked_, which is changing the state of their application. It would be better, in this world without an AST, to leave the mutations in place. Removing the mutations causes more issues than it tries to solve.

## Future solve

I think we should really explore transforming their DOM into an AST and modifying it after we've `POST`'ed it to the agent process. This would 100% ensure we're not mutating the customers DOM in any way. This is not a simple change, but I believe it is the _right_ change. 

## TODO
- [x] Fix the tests that fail related to DOM clean up

### DIFF
--- a/test/integration/agent-integration.test.ts
+++ b/test/integration/agent-integration.test.ts
@@ -67,8 +67,8 @@ describe('Integration test', () => {
       const domSnapshot = await snapshot(page, 'Buildkite snapshot')
       expect(domSnapshot).contains('Buildkite')
     })
-  })
 
+  })
   describe('on local test cases', () => {
     const testCaseDir = `${__dirname}/testcases`
     const PORT = 8000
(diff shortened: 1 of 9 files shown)


test/integration/agent-integration.test.ts
```diff
@@ -96,33 +96,36 @@ describe('Integration test', () => {
...
+      })
+
+      it('serializes the CSSOM', async () => {
+        const domSnapshot = await snapshot(page, 'Serialize CSSOM')
+
+        expect(domSnapshot).to.contain('data-percy-cssom-serialized')
+        expect(domSnapshot).to.contain('.box { height: 500px; width: 500px; background-color: green; }')
+
+        // we want to ensure mutiple snapshots are successful
+        const secondDomSnapshot = await snapshot(page, 'Serialize CSSOM twice')
```
maprules1000: 👍

PR #167, opened 2019-04-25
Fix serializing inputs

blake-newman:
- Move logic into self contained functions to help describe the
procedures of serializing inputs
- Add integration tests capturing every case for reseting the dom after
serialisation


blake-newman:
@Robdel12 TIL, textarea does have `type`.

blake-newman:
@Robdel12 @djones 

I have updated this PR. I have added 16 tests to capture all the variations to ensure that serialising is cleaned up.

djones:
We're putting this PR on hold as it will be superseded by https://github.com/percy/percy-agent/pull/208. Thanks for pushing the conversation here @blake-newman 👍

PR #168, opened 2019-04-25
feature: Support parallel snapshots calls

blake-newman:
Fix parallel builds

When parallel builds occur there is the posibility that the same
assets will be triggerd on the same pooled page, this means that there
is a likely outcome that the resource will be intercepted and handled
twice, this is illegal with puppeteer.

Polling available pages up to 10 is also memory intensive for sequential
builds as 10 pages will be in memory. This can cause issues in low
memory enviroments.

- Remove pooling and open close pages on demand to reduce memory
footprint
- Open new page on demand to ensure that parallel builds that request
same resource do not conflict which causes the puppeteer to throw up
error that request is already handled.
- Add catch clauses around request interception to ensure that e2e tests
continue if there is an issue with percy.


Robdel12:
This looks _really_ good to me 🔥. Excited about these changes and I like the approach. 

@djones any thoughts on this?

djones:
Thanks for another great PR @blake-newman! It's delightful to see someone who doesn't work at Percy go this deep.

Just to give you a heads up, the asset discovery flow was designed this way with performance in mind. There's a 75ms or so cost to opening a page in Chromium, so the current flow keeps open 10 pages and opens them when the agent is started. This means the time to take a single snapshot doesn't need to include opening a new page in the browser.

You're quite right to point out that in doing so, we have broken parallel snapshotting and we would definitely like to fix that. This issue https://github.com/percy/percy-puppeteer/issues/52 is highly relevant and this PR would likely fix it.

I want to measure what performance impact this has on a per-snapshot level and see if that's something we can stomach if it's slower. We have fairly good profiling if you run `@percy/agent` with `LOG_LEVEL=debug` set. It outputs a bunch of timings. The one of particular interested here is `assetDiscoveryService.discoverResources`.

One thing we have been considering is https://github.com/thomasdondorf/puppeteer-cluster it would allow us to have a pool of Chromium instances running and warmed up. Something like this would be one way to keep per-snapshot performance high and have parallel snapshotting working (though it would consume more memory).

blake-newman:
I had thought that above may be the case, 75ms seems to be quite high at will certainly have a knock-on effect for sequential builds. 

We have also found that the pool can consume a large amount of memory causing issues when running in parallel as there is a large demand for memory which causes chromium to crash. 

Based on the above information, I believe that a partial but adequate solution to fix parallel builds would be to run asset discovery in two formats depending on how the tests are running. As a sequential test run has more access to memory as there is less demand so pooling makes sense.

However, when running in parallel you have less memory available thus the pool can use up additional unavailable memory which becomes a bottleneck and can cause crashes. When running in parallel cost of the setup of opening a new page becomes less of an issue because parallel as a whole reduces test runs significantly. 

I will submit an update to give a best of both worlds based on the test environment, as I feel enabling parallel builds is a huge win and optimizations to make snapshotting for those builds faster can be done at a later stage. However, correctly put we should ensure that sequential builds are not deopmitized when taking snapshots.

blake-newman:
@djones have pushed up new changes and believe this solution covers both use cases, making builds optimized for the resource they require most. 

In the case of parallel there is less resource for memory so not using a pool makes sense as it's memory heavy. The is also more bandwidth for execution time since parallel comes at a huge speed boost by itself. Further optimisation for future would be to create a revolving pool that spins up clean tabs after use to ensure that screen shots can execute in a quicker manner, however that comes at a cost for memory and with builds already crashing with 16gb of memory with a pool (believe this is due to high memory usage in cypress causing percy to push it to the limit) this would need some more investigation to ensure that test runs are stable on low memory environments. 

For serial there is less resource for processing power but memory is more readily available for use, so having a pool helps in the case of serial test suites.

blake-newman:
@Robdel12 @djones sorry to push on this, wondering if you have had chance to look?

djones:
Hi @blake-newman, Sorry about the delay... this is super important to get right as you can imagine.

This is definitely getting closer. I'm leaning towards killing off `isParallel` and just having that on by default. Opening up to 10 pages really isn't that expensive. Would like to pick a good default for everyone.

src/services/asset-discovery-service.ts
```diff
@@ -124,29 +152,40 @@ export default class AssetDiscoveryService extends PercyClientService {
...
+    pooledPage: puppeteer.Page | null,
+    width: number,
+    domSnapshot: string,
+    rootResourceUrl: string,
+    enableJavaScript: boolean,
+  ): Promise<any[]> {
     logger.debug(`discovering assets for width: ${width}`)
-
+  
+    const page = pooledPage || await this.createPage(browser)
```
djones: If the `pooledPage` doesn't exist, we create one here. But that newly created page is not added to `this.pages` so I don't think it will ever get cleaned up.

src/services/asset-discovery-service.ts
```diff
@@ -79,8 +105,10 @@ export default class AssetDiscoveryService extends PercyClientService {
     profile('--> assetDiscoveryService.discoverForWidths', {url: rootResourceUrl})
     const resourcePromises: Array<Promise<any[]>> = []
     for (let idx = 0; idx < widths.length; idx++) {
+      const page = this.pages !== null ? this.pages[idx] : null
```
djones: I think this could be problematic.

Let's say we have two snapshot requests that come in. Each with 7 widths. We'll assume we have `isParallel` set to `true`.

* Before any snapshots come in, we've created 10 pages.
* The first snapshot request comes in and the first 7 pages are allocated
* While that begins, a second snapshot request comes in. It looks like we'll reallocate those same 7 pages even though they might be in use?

djones: A resource pool would definitely resolve these issues.

What you would do is specify the upper bounds of your pool... I actually think we could make it just 5 pages by default because the vast majority of folks take snapshots at 2-3 widths only (and no parallel requests). We could potentially make the max pool size configurable with a command line flag so people could control the max resources we'll burst up to.

And then if the pool is full, we'll have to wait until a page becomes free to use. And that mechanism will be how you can handle parallel snapshot requests. Worst case, you wait a little bit.

I did a bit of Googling around. https://github.com/coopernurse/node-pool was the first thing that came up. There might be other options out there too.

Would be curious to get your thoughts on all this @blake-newman

blake-newman: I believe you are right, resolving this way would mean that there could be a clash, thus this is the exact problem with parallel tests currently. Albeit this logic is now only available in serial execution, thus making hitting the same page in the pool near (probably) impossible. 

The main issue trying to be solved is ensuring that there is no clash with requests on the same page, which with a pool it likely to be observed. IMO I believe this solution does justice to enabling parallel builds with the minimal amount of changes, to prevent likely bugs. However, I don't believe it is the best solution, as a revolving pool of pages will ensure optimised run times for parallel builds. 

By the nature of parallel builds running a magnitude faster, i believe we should aim to support parallel builds at a minimum. For example we have a 5 mins test suite, enabling parallel builds on our CI will bring the results down to around 1-2mins which is a huge win in a CI environment. In this case the changes here would need to introduce roughly 3 mins of time. Based on 75ms load time for a page this would mean 2400 snapshots in the whole test run (:money_mouth_face:), thus i don't think it's a huge problem to load on demand. 

Obviously, i'm happy to work on a full solution that both suits sequential and parallel. I'm a big fan of delivering smaller change sets to enable value, i would be concerned that introducing a more complex solution to enable parallel builds will cause unexpected bugs that we can't forsee. However, this is your codebase so happy to do either to get this in to improve the lives for every engineer. 

Tackling the points raised, i believe the `node-pool` which i also looked into would be a great solution. Although i think it would require some large code changes to ensure we cope with scheduling.

src/services/asset-discovery-service.ts
blake-newman: Yes, we don't add it to the pool. As it is cleaned up as soon as it is done with: https://github.com/percy/percy-agent/pull/168/files/11b1bd8df53dc0abc0d03bf0dc00fca8999978be#diff-81e4b6d82b5aac3d228d58c084bdc537R225

src/services/asset-discovery-service.ts
blake-newman: @djones i have pushed changes to use `generic-pool` as i found it pretty simple to implement. 

The question now is how big we make the pool. Memory issues we was facing was not a problem with percy but due to this issue: https://github.com/cypress-io/cypress/issues/350

So i believe setting to `min: 2, and max: 10` should be sufficient. However a larger pool will probably improve performance with this solution as there will be more pages to use at any given time. So for parallel builds this will be more optimised. Perhaps giving the option to set max pool size via `PERCY_POOL_SIZE` will allow configuration for low / high memory environments.

blake-newman:
I am currently creating a test suite capturing all these changes along with `#167` to validate all the changes work as expected. (So far look positive)

src/services/asset-discovery-service.ts
```diff
@@ -12,14 +13,12 @@ interface AssetDiscoveryOptions {
 export default class AssetDiscoveryService extends PercyClientService {
   responseService: ResponseService
   browser: puppeteer.Browser | null
-  pages: puppeteer.Page[] | null
+  pool: pool.Pool<puppeteer.Page> | null
```
djones: perhaps `pool` => `pagePool`

src/services/asset-discovery-service.ts
```diff
@@ -12,14 +13,12 @@ interface AssetDiscoveryOptions {
...
 
   readonly DEFAULT_NETWORK_IDLE_TIMEOUT: number = 50 // ms
   networkIdleTimeout: number // ms
 
-  // How many 'pages' (i.e. tabs) we'll keep around.
-  // We will only be able to process at most these many snapshot widths.
-  readonly PAGE_POOL_SIZE: number = 10
+  readonly MAX_SNAPSHOT_WIDTHS: number = 10
```
djones: how about:

```
readonly PAGE_POOL_SIZE_MIN_DEFAULT: number = 2
readonly PAGE_POOL_SIZE_MAX_DEFAULT: number = 10
```

src/services/asset-discovery-service.ts
```diff
@@ -29,12 +28,21 @@ export default class AssetDiscoveryService extends PercyClientService {
...
-    this.pages = null
+    this.pool = null
   }
 
   async setup() {
+    profile('-> assetDiscoveryService.setup')
+    const browser = this.browser = await this.createBrowser()
+    this.pool = await this.createPagePool(() => {
+      return this.createPage(browser)
+    }, this.DEFAULT_WIDTHS.length, this.MAX_SNAPSHOT_WIDTHS)
```
djones:
```suggestion
    }, this. PAGE_POOL_SIZE_MIN_DEFAULT, this. PAGE_POOL_SIZE_MAX_DEFAULT)
```

src/services/asset-discovery-service.ts
```diff
@@ -29,12 +28,21 @@ export default class AssetDiscoveryService extends PercyClientService {
     this.responseService = new ResponseService(buildId)
     this.networkIdleTimeout = options.networkIdleTimeout || this.DEFAULT_NETWORK_IDLE_TIMEOUT
     this.browser = null
-    this.pages = null
+    this.pool = null
```
djones:
```suggestion
    this.pagePool = null
```

src/services/asset-discovery-service.ts
```diff
@@ -29,12 +28,21 @@ export default class AssetDiscoveryService extends PercyClientService {
...
   }
 
   async setup() {
+    profile('-> assetDiscoveryService.setup')
+    const browser = this.browser = await this.createBrowser()
+    this.pool = await this.createPagePool(() => {
```
djones:
```suggestion
    this.pagePool = await this.createPagePool(() => {
```

src/services/asset-discovery-service.ts
```diff
@@ -44,28 +52,48 @@ export default class AssetDiscoveryService extends PercyClientService {
...
-        return page.setRequestInterception(true).then(() => page)
-      })
-      pagePromises.push(promise)
-    }
-    this.pages = await Promise.all(pagePromises)
-    profile('-> assetDiscoveryService.browser.newPagePool')
+    return browser
+  }
+
+  async createPagePool(exec: () => PromiseLike<puppeteer.Page>, min: number, max: number) {
```
djones: nice

src/services/asset-discovery-service.ts
```diff
@@ -44,28 +52,48 @@ export default class AssetDiscoveryService extends PercyClientService {
...
-    if (!this.browser || !this.pages || !this.pages.length) {
-      logger.error('Puppeteer failed to open with a page pool.')
+    if (this.browser === null) {
+      logger.error('Puppeteer failed to open browser.')
       return []
     }
 
-    if (options.widths && options.widths.length > this.pages.length) {
-      logger.error(`Too many widths requested. Max allowed is ${this.PAGE_POOL_SIZE}. Requested: ${options.widths}`)
+    if (!this.pool) {
```
djones:
```suggestion
    if (!this.pagePool) {
```

src/services/asset-discovery-service.ts
```diff
@@ -124,33 +151,45 @@ export default class AssetDiscoveryService extends PercyClientService {
   }
 
   async teardown() {
-    await this.closePages()
+    await this.cleanPool()
```
djones:
```suggestion
    await this.cleanPagePool()
```

src/services/asset-discovery-service.ts
```diff
@@ -124,33 +151,45 @@ export default class AssetDiscoveryService extends PercyClientService {
...
+  private async resourcesForWidth(
+    pool: pool.Pool<puppeteer.Page>,
+    width: number,
+    domSnapshot: string,
+    rootResourceUrl: string,
+    enableJavaScript: boolean,
+  ): Promise<any[]> {
     logger.debug(`discovering assets for width: ${width}`)
 
+    profile('--> assetDiscoveryService.pool.acquire', {url: rootResourceUrl})
```
djones: Very very nice. If I read this correctly, we'll get timing information on when we actually do have to wait for a page in the pool to become available.

src/services/asset-discovery-service.ts
```diff
@@ -174,31 +213,44 @@ export default class AssetDiscoveryService extends PercyClientService {
...
 
-    page.removeAllListeners()
+    return []
+  }
 
-    profile('--> assetDiscoveryServer.waitForResourceProcessing')
-    const maybeResources: any[] = await Promise.all(maybeResourcePromises)
-    profile('--> assetDiscoveryServer.waitForResourceProcessing')
-    return maybeResources.filter((maybeResource) => maybeResource != null)
+  private async cleanPool() {
```
djones:
```suggestion
  private async cleanPagePool() {
```

src/services/asset-discovery-service.ts
```diff
@@ -174,31 +213,44 @@ export default class AssetDiscoveryService extends PercyClientService {
...
+    if (this.pool === null) { return }
```
djones:
```suggestion
    if (this.pagePool === null) { return }
```

src/services/asset-discovery-service.ts
```diff
@@ -174,31 +213,44 @@ export default class AssetDiscoveryService extends PercyClientService {
...
+    await this.pool.drain()
```
djones:
```
    await this.pagePool.drain()
```

src/services/asset-discovery-service.ts
```diff
@@ -174,31 +213,44 @@ export default class AssetDiscoveryService extends PercyClientService {
...
+    await this.pool.clear()
```
djones:
```suggestion
    await this.pagePool.clear()
```

src/services/asset-discovery-service.ts
```diff
@@ -174,31 +213,44 @@ export default class AssetDiscoveryService extends PercyClientService {
...
+    this.pool = null
```
djones:
```
    this.pagePool = null
```

test/services/snapshot-service.test.ts
```diff
@@ -37,6 +39,21 @@ describe('SnapshotService', () => {
       expect(snapshotResponse.body).to.deep.equal({data: {id: snapshotId}})
       expect(snapshotResponse.statusCode).to.eq(200)
     })
+
+    it('creates multiple snapshots', async () => {
```
djones: I can confirm on master, this test passes when `length: 1` and fails with `length: 2` or more.

src/services/asset-discovery-service.ts
```diff
@@ -9,17 +10,19 @@ interface AssetDiscoveryOptions {
   networkIdleTimeout?: number
 }
 
+const DEFAULT_PAGE_POOL_SIZE = process.env.PERCY_POOL_SIZE
```
djones: we'll shift this to use the `.percy.yml` config file later. Happy to leave this is for now, but don't rely on this

src/services/asset-discovery-service.ts
```diff
@@ -124,33 +155,45 @@ export default class AssetDiscoveryService extends PercyClientService {
...
+    const page = await pool.acquire()
+    profile('--> assetDiscoveryService.pool.acquire')
     await page.setJavaScriptEnabled(enableJavaScript)
     await page.setViewport(Object.assign(page.viewport(), {width}))
 
     page.on('request', async (request) => {
-      if (!this.shouldRequestResolve(request)) {
-        await request.abort()
-        return
+      try {
```
Robdel12: 😍

src/services/asset-discovery-service.ts
```diff
@@ -174,31 +217,44 @@ export default class AssetDiscoveryService extends PercyClientService {
...
+
+      profile('--> assetDiscoveryService.pool.release', {url: rootResourceUrl})
+      await page.removeAllListeners('request')
+      await page.removeAllListeners('requestfinished')
+      await page.removeAllListeners('requestfailed')
+      await pool.release(page)
+      profile('--> assetDiscoveryService.pool.release')
+
+      return maybeResources.filter((maybeResource) => maybeResource != null)
+    } catch (error) {
```
Robdel12: Love that we're going to catch these async errors now. Does it get rid of all of the deprecations?

PR #177, opened 2019-05-02
fix: Revert change to module bundling

Robdel12:
Looks like `iife` from rollup isn't exposing agent as a global. We
need `window.PercyAgent =` not `var PercyAgent=`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3508,12 +3508,6 @@
         "vary": "^1"
       }
     },
-    "corser": {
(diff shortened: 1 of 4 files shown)


Robdel12:
See: https://github.com/jfhbrook/node-ecstatic/issues/255 for the last commit. Once that's resolved we can add `http-server` back and uncomment those tests :/

thornjad:
jfhbrook/node-ecstatic#256 will fix http-server

PR #200, opened 2019-05-14
fix: Improved readme and help text

timhaines:
Reformats some wording, adds context around finalizing a build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ USAGE
 
 ## `percy exec`
 
-Start and stop Percy around a supplied command.
+Start and stop Percy around a supplied command
 
 ```
 USAGE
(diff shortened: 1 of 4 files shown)


djones:
@timhaines that bit of the docs is autogenerated by Oclif. To update the wording, you'll want to update the description for each command, e.g. https://github.com/percy/percy-agent/blob/master/src/commands/exec.ts#L7

This has an added bonus that you'll be fixing not only the readme but the help text for the CLI 😄

Note the readme won't autogenerate until the release flow is run and that only happens when a new version is shipped. Technically I think this PR title should start with `docs:` and that _won't_ ship a new version. If you feel strongly that you want to ship this update to the readme soon, feel free to change this PR to have `fix:` at the front.

Side note: in the [raw readme](https://raw.githubusercontent.com/percy/percy-agent/master/README.md) there's some magic comments like 

```markdown
<!-- commands -->
Command docs are generated into this space.
<!-- commandsstop -->
```

and so that's how the auto doc generation works.

PR #201, opened 2019-05-15
build: Use `npm publish` straight up for dual release

djones:
Dual releasing in #196 did not end up actually working. This was because `semantic-release` recognized there was already a git tag for this version and decided there was nothing further to do.

The logs look like this:

```
[7:33:24 PM] [semantic-release] › ℹ  Found git tag v0.4.8 associated with version 0.4.8
[7:33:24 PM] [semantic-release] › ℹ  Found 0 commits since last release
[7:33:24 PM] [semantic-release] › ℹ  Start step "analyzeCommits" of plugin "@semantic-release/commit-analyzer"
[7:33:24 PM] [semantic-release] [@semantic-release/commit-analyzer] › ℹ  Analysis of 0 commits complete: no release
[7:33:24 PM] [semantic-release] › ✔  Completed step "analyzeCommits" of plugin "@semantic-release/commit-analyzer"
[7:33:24 PM] [semantic-release] › ℹ  There are no relevant changes, so no new version is released.
```

I had a look into `semantic-release` and didn't find any decent way of getting it to work so I'm swapping the second release command here to be a straight up `npm publish`.

We avoid attempting to republish an existing version by checking with npm first.


.circleci/config.yml
```diff
@@ -36,16 +36,21 @@ jobs:
       - checkout
       - run: npm ci
       - run: npx semantic-release
-      - run: sed -i 's/@percy\/agent/percy/g' package.json
-      - run: npm ci
-      - run: npx semantic-release
+      - run: |
+          sed -i 's/@percy\/agent/percy/g' package.json
+          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
+          npm publish
```
Robdel12: Will this fail CI when it tries to publish over the same version?

> What this means is that npm publish -f will not work. The first step, where it deletes the old copy, will work. But then trying to publish over it will fail with a 403 Forbidden response. You’ll have to change the version number to make it work.

https://blog.npmjs.org/post/77758351673/no-more-npm-publish-f

djones: You're absolutely right. I reran a build that was green and now it's red https://circleci.com/gh/percy/percy-agent/1970

I'll find a workaround for that.

Robdel12: This is the little node script we used to release packages: https://github.com/bigtestjs/meta/blob/master/bin/release

Basically checks the package.json version in the npm registry and if it's not present, go ahead and release.

djones: @Robdel12 ready for another look

PR #207, opened 2019-05-23
test: Setup Karma over Testem

Robdel12:
## What is this?

This will allow us to run our tests in an iframe (so it's isolated from any test output), have a warmed cache & watch server while writing tests. Also cleans up some of the TODOs that were present in the old `index.html` file.

I was spending too much time killing and restarting the test command while trying to write tests for the DOM refactor. Also it was really difficult to work around the fact that the tests were running in the same DOM that the test output was running in. 

I removed two test files. We still have coverage for it in our integration tests and with my refactor of how we handle the DOM we'll have much more robust tests. It wasn't worth trying to fix them for this PR.


test/percy-agent-client/karma.conf.js
```diff
@@ -0,0 +1,45 @@
...
+      // tells Karma that we'll be using Webpack to process this file
+      "./index.js": ["webpack"]
+    },
+
+    // Mocha reporter options
+    mochaReporter: {
+      showDiff: true
+    },
+
+    // our webpack config exports a function
```
Robdel12: This comment needs to be updated

Issue #188, opened 2019-05-09
CSSOM serialisation issue


blake-newman:
Sorry for this vague issue report however I'm not experienced on the purpose of CSSOM serialising. However, during #167 i have noticed an issue with CSSOM serialisation causing multiple snapshots on a page to slow by 2x sequentially per test due to the serialising. 

When running a tests the in the PR it is now much more evident because there is lots more tests ran. 

For each test this [css](https://github.com/percy/percy-agent/blob/master/test/percy-agent-client/test.html#L35) get duplicated. 

`.cls-1 { fill: rgb(141, 123, 216); }` becomes `.cls-1 { fill: rgb(141, 123, 216); }.cls-1 { fill: rgb(141, 123, 216); }` .... and so forth. By the 16th test this is repeated 65536 times.

I assume this may affect actual test speeds, as per snapshot would cause the same thus why raising as an issue. 
 
I can provide more details if needed.

Robdel12:
Hey Blake! Yea this is something we’re actively working on at the moment (though I’m on vacation right now). 

Thanks for the report on this, I’ll make sure this makes it into the refactoring that’s taking place

blake-newman:
Thanks for the response, this was mainly to make sure you were aware of the issue.

Enjoy your vacation! 🏖️

PR #208, opened 2019-05-23
fix: Serialize in-memory attributes into deep DOM clone

Robdel12:
## Small background

Percy Agent is the underlying package that powers most of our SDKs now. It provides a CLI to handle creating the builds (among other features), a node process for processing the DOM, a way to capture assets, and provides the JS we use to actually capture the DOM from the customer's browser.

There is some state within the page that we need to capture but aren't present in the DOM we take from the browser (form element values, CSS Object Model, Canvas elements, etc).

In order for us to properly capture all of the state within the customer's browser, we need to serialize those values into the DOM we are capturing.

## The problem

At this time we're modifying the DOM within the customer's browser to capture that state. This can be harmful to their applications state. You can see how this works now here: https://github.com/percy/percy-agent/blob/593d82a24cacbf9058eb8c0a0c301d967f40ad38/src/percy-agent-client/percy-agent.ts#L58-L95

First, we take the DOM that's passed from the snapshot method and serialize those in-memory values into the customers DOM. Then we take a clone of it after we've modified it.

There are two main problems with the current approach:

- It modifies the DOM within their app, which can influence their test suite and applications state.
- For CSSOM specifically, it breaks the way the app handles styles as soon as we serialize the styles into the ownerNode

## The fix

We can solve the two issues listed above by cloning the DOM that's passed from the `snapshot` method as soon as we get it and then serializing the in-memory values into the DOM clone. This would require us to change how the current system works by passing both the original DOM and the clone that we've taken. 

#### Isolate DOM transforms into a single class

This PR isolates all of the DOM transformations we make to a single DOM class & pulls the two different `serialize-*.ts` files into that class. This will make it easy to contain, manage, and document each DOM transformation that happens. The steps that we take to capture this information are a little confusing and nuanced. I’d like to make it as clear as possible the type of transforms that are happening. 

This will also make testing MUCH easier. It will also make it possible for us to store any future errors we want to log on the class, and then pass that class along as debug output. 

#### Inputs

To capture the values from the inputs we will still need to modify their DOM but it's only adding a Percy data attribute with a guid.  This is because we need to be able to locate the various inputs in the original DOM and serialize them into the DOM clone. We cannot rely on their selectors or attributes. It also must be done before the clone is taken so the guid is present and the consistent in both DOM trees. 

I believe this is as non-invasive as we can get with this approach.

#### CSSOM

There are two small changes to get stable CSSOM serialization across snapshots:

- Serialize the CSSOM from the original document into the DOM clone
- _Create_ a style tag and append it to the same `parentNode` of the CSSOM’s `ownerNode`. We **do not** want to serialize these styles into the CSSOM’s `ownerNode` because it will break how the various libraries modify the CSSOM. 

We have [learned in the past that removing styles](https://github.com/percy/percy-agent/pull/160) that were serialized into the CSSOM’s `ownerNode` will call `deleteRule()` in the CSSOM, which removes all styles present in the application (while also breaking the library that generates the styles). We want to avoid adding or changing **any** styles that the CSSOM owns. 

#### Test coverage

This needs to be thoroughly tested. We should make sure:

- We only modify the DOM to add the guid data attrs to form elements
- We properly capture the CSSOM (in many snapshots, since that’s been an issue. [Good repo here](https://github.com/montezume/percy-puppeteer-emotion/tree/ml-try-prod)) (Here's a branch with this version of agent running that fixes the issues: https://github.com/Robdel12/percy-puppeteer-emotion/tree/ml-try-prod)
    - We do not serialize the CSSOM when JS is enabled (since that will be generated when the snapshot is rendered in our env)
- We properly capture input values 
- CSSOM is not broken by adding more rules after clone
- `snapshotString()` always returns a string with a doctype
- `domTransformation` works as expected
- `domTransformation` errors are captured and do not error the whole suite

## Open TODOs

I'm not sure any of these are blocking, but could be convinced they are.

- [x] Add better typings than `any`. 
- [ ] Append the new style tag to the same `parentNode` as the CSSOM's style `ownerNode`


src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,205 @@
...
+    // the full page. Using a dom transformation is how you'd acheive that.
+    if (this.options.domTransformation) {
+      try {
+        dom = this.options.domTransformation(dom)
+      } catch (error) {
+        console.error('Could not transform the dom: ', error)
+      }
+    }
+
+    const snapshotString = doctype + dom.outerHTML
```
Robdel12: Probably an unneeded var here

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * Capture in-memory styles & serialize those styles into the cloned DOM.
+   *
+   * Without this, applications that use packages like `styled-components` will be
+   * missing styles and appear broken. The CSSOM provides a way to create CSS
+   * that only lives in memory (not an asset or in the DOM).
+   *
+   */
+  private serializeCssOm(documentClone: any) {
```
Robdel12: It'd be nice to do some perf testing here (re #188). I _think_ we're going to have to bite the perf cost here of iterating through each in-memory stylesheet each time. 

Maybe in the future we can do some caching. Feels dangerous though, I'd rather capture the CSSOM each time to be 100% sure it's correct.

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,205 @@
...
+        const cssRules = Array.from(styleSheet.cssRules)
+        const serializedStyles = cssRules.reduce((prev: string, cssRule: any) => {
+          return prev + cssRule.cssText
+        }, '')
+
+        // Append the serialized styles to the cloned document
+        $style.type = 'text/css'
+        $style.setAttribute('data-percy-cssom-serialized', 'true')
+        $style.innerHTML = serializedStyles
+        // TODO, it'd be better if we appended it right after the ownerNode in the clone
```
Robdel12: I think this is something that's nice to have, but not critical for this PR to get through.

test/integration/agent-integration.test.ts
```diff
@@ -112,10 +113,12 @@ describe('Integration test', () => {
         await page.click('#testRadioButton')
 
         const domSnapshot = await snapshot(page, 'Serialize input elements')
-        expect(domSnapshot).to.contain('test input value')
-        expect(domSnapshot).to.contain('type="checkbox" checked')
-        expect(domSnapshot).to.contain('type="radio" checked')
-        expect(domSnapshot).to.contain('test textarea value')
+        const $ = cheerio.load(domSnapshot)
```
Robdel12: [Cheerio](https://github.com/cheeriojs/cheerio) makes working with the _string_ of the DOM much easier by giving a jQuery API to transverse the string. This was needed because the `contain` method checks to make sure that _exact_ string matches. So the order of `type="radio" checked` needs to be exact (which it isn't) to pass.

Cheerio allows us to query for that element in the DOM and check its attributes.

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      // @ts-ignore
+      expect(document.querySelector('.delete-me').innerText).to.equal('Delete me')
+    })
+
+    // it's possible the user provides code that errors when we execute it
+    it('gracefully catches errors', () => {
+      expect(dom.snapshotString()).to.not.contain('Delete me', 'delete-me')
+      // invoke the transform function again to try and remove a non-existent element
+      expect(dom.snapshotString()).to.contain('Hello DOM testing')
+      expect(consoleStub.calledOnce).to.equal(true)
```
Robdel12: Add an expect before the first `snapshotString` method call to ensure `console.error` has not been called.

src/percy-agent-client/dom.ts
djones: Yeah agreed. Totally fine to just shove this into the return value directly.

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Serialize parts of the DOM that aren't preserved in a deep clone. Anything
+   * that is not encoded in the DOM tree needs to be accounted for here (CSSOM,
+   * input values, canvas, etc).
+   *
+   * This method should always capture these values from the origial DOM and
+   * serialize into the cloned DOM. Never mutate the origial DOM.
+   *
+   */
+  private stabilizeDOM(clonedDOM: any): HTMLElement {
```
djones:
```suggestion
  private stabilizeDOM(clonedDOM: HTMLDocument): HTMLElement {
```

src/percy-agent-client/dom.ts
djones:
```suggestion
  private serializeCssOm(documentClone: HTMLDocument) {
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,205 @@
+/**
+ * A single class to encapsulate all DOM operations that need to be preformed to
+ * capture the customers application state.
+ *
+ */
+class DOM {
+  originalDOM: any
```
djones:
```suggestion
  originalDOM: HTMLDocument
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,205 @@
...
+  clonedDOM: any
```
djones:
```suggestion
  clonedDOM: HTMLDocument
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
+export interface DOMOptions {
+  enableJavaScript?: boolean,
+  domTransformation?: (dom: any) => void
+}
+
+/**
+ * A single class to encapsulate all DOM operations that need to be preformed to
```
timhaines: performed

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+ * capture the customers application state.
```
timhaines: customer's

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+   *
+   */
+  snapshotString(): string {
+    let dom = this.clonedDOM
+    const doctype = this.getDoctype()
+
+    // Sometimes you'll want to transform the DOM provided into one ready for snapshotting
+    // For example, if your test suite runs tests in an element inside a page that
+    // lists all yours tests. You'll want to "hoist" the contents of the testing container to be
+    // the full page. Using a dom transformation is how you'd acheive that.
```
timhaines: achieve

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    return clone
+  }
+
+  /**
+   * Serialize parts of the DOM that aren't preserved in a deep clone. Anything
+   * that is not encoded in the DOM tree needs to be accounted for here (CSSOM,
+   * input values, canvas, etc).
+   *
+   * This method should always capture these values from the origial DOM and
```
timhaines: original

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+   * serialize into the cloned DOM. Never mutate the origial DOM.
```
timhaines: original - starting to wonder if this is a keyboard issue on the letter 'n' :)

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+    this.clonedDOM = this.cloneDOM()
+  }
+
+  /**
+   * Returns the final DOM string with all of the necessary transforms
+   * applied. This is the string that is passed to the API and then rendered by
+   * our API.
+   *
+   */
+  snapshotString(): string {
```
timhaines: I'd wondering about being more specific here `domSnapshotString` or some other name - to avoid possible confusion with Percy's Snapshots.  I.e. in the near future we think each Percy Snapshot may have multiple DOM Snapshots.

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+   * that is not encoded in the DOM tree needs to be accounted for here (CSSOM,
+   * input values, canvas, etc).
+   *
+   * This method should always capture these values from the origial DOM and
+   * serialize into the cloned DOM. Never mutate the origial DOM.
+   *
+   */
+  private stabilizeDOM(clonedDOM: any): HTMLElement {
+    let stabilizedDOMClone
+    stabilizedDOMClone = this.serializeInputElements(clonedDOM)
```
timhaines: does it make sense to combine this with the line above?

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    // We only want to serialize the CSSOM if JS isn't enabled.
+    if (!this.options.enableJavaScript) {
+      stabilizedDOMClone = this.serializeCssOm(clonedDOM)
```
timhaines: I think this will overwrite the `stabilizedDOMClone` with input elements stablized?

Robdel12: You're right. I should write a test to make sure this works together.

src/percy-agent-client/dom.ts
Robdel12: You would be right! And spacebar and p 😂

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  /**
+   * Capture in-memory styles & serialize those styles into the cloned DOM.
+   *
+   * Without this, applications that use packages like `styled-components` will be
+   * missing styles and appear broken. The CSSOM provides a way to create CSS
+   * that only lives in memory (not an asset or in the DOM).
+   *
+   */
+  private serializeCssOm(documentClone: HTMLDocument) {
```
timhaines: Not particularly fussed, but expected this to be capitalized like this: `serializeCSSOM`

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+        // TODO, it'd be better if we appended it right after the ownerNode in the clone
+        documentClone.head.appendChild($style)
+      }
+    })
+
+    return documentClone
+  }
+
+  /**
+   * A single place to mutate the origial DOM. This should be the last resort!
```
timhaines: original

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+   * This will change the customers DOM and have a possible impact on the
+   * customers application.
```
timhaines: customer's

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+   *
+   */
```
timhaines: May also want to include comments above saying currently we're only using it to add data-percy-element-id's to inputs which should be safe.  Also that this mutation can apply to the same original DOM multiple times within one test/snapshot set.  This can be dangerous, which is why we suggest strictly minimizing changes made here.

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+        Math.random()
+          .toString(36)
+          .substr(2, 9)
+
+      $el.setAttribute('data-percy-element-id', ID)
+    }
+
+    const formNodes = this.originalDOM.querySelectorAll('input, textarea')
+    const formElements = Array.from(formNodes)
+    // loop through each for element and apply an ID for serialization later
```
timhaines: form

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,210 @@
...
+   */
+  private mutateOriginalDOM() {
+    function createUID($el: Element) {
+      const ID =
+        '_' +
+        Math.random()
+          .toString(36)
+          .substr(2, 9)
+
+      $el.setAttribute('data-percy-element-id', ID)
```
timhaines: I guess it's safe to set these to new values even if they already existed?  i.e. a second snapshot on the same dom will replace the ids added by the first snapshot..

Robdel12: We could probably save some work for ourselves by checking if it exists first. It's not too expensive to do it each time but I think it's worth checking first

src/percy-agent-client/dom.ts
Robdel12: I like the name `domSnapshotString`.

Though, it is on the `DOM` class:

```js
let dom = new DOM(document);

// this
dom.snapshotString()
// vs
dom.domSnapshotString()
```

I don't have a strong feel either way

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * Capture in-memory styles & serialize those styles into the cloned DOM.
+   *
+   * Without this, applications that use packages like `styled-components` will be
+   * missing styles and appear broken. The CSSOM provides a way to create CSS
+   * that only lives in memory (not an asset or in the DOM).
+   *
+   */
+  private serializeCSSOM(documentClone: HTMLDocument) {
+    const styleSheets = Array.from(this.originalDOM.styleSheets)
```
djones:
```suggestion
const styleSheets = Array.from(this.originalDOM.styleSheets) as CSSStyleSheet[]
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    styleSheets.forEach((styleSheet: any) => {
```
djones:
```suggestion
    styleSheets.forEach((styleSheet) => {
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+        const ownerNode = styleSheet.ownerNode as HTMLElement
+        const hasStyleInDom =
+          ownerNode.innerText && ownerNode.innerText.length > 0
+
+        return !hasHref && !hasStyleInDom && styleSheet.cssRules
+      }
+
+      if (isCSSOM()) {
+        const $style = documentClone.createElement('style')
+        const cssRules = Array.from(styleSheet.cssRules)
```
djones:
```suggestion
        const cssRules = Array.from(styleSheet.cssRules) as CSSRule[]
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+        const serializedStyles = cssRules.reduce((prev: string, cssRule: any) => {
```
djones:
```suggestion
        const serializedStyles = cssRules.reduce((prev: string, cssRule) => {
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+   * respective elements in the cloned DOM.
+   *
+   * Form element values for non-controlled elements won't be captured with a
+   * domClone. To fix this, we explicitly set the `value` attribute on those
+   * form controls to make sure they persist in snapshots.
+   *
+   */
+  private serializeInputElements(clonedDOM: HTMLDocument): HTMLDocument {
+    const formNodes = this.originalDOM.querySelectorAll('input, textarea')
+    const formElements = Array.from(formNodes)
```
djones:
```suggestion
    const formElements = Array.from(formNodes) as HTMLFormElement[]
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    formElements.forEach((elem: any) => {
```
djones:
```suggestion
    formElements.forEach((elem) => {
```

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+      const inputId = elem.attributes['data-percy-element-id'].value
```
djones:
```suggestion
      const inputId = elem.getAttribute('data-percy-element-id')
```

I think this is comparable?

test/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "../tsconfig",
   "compilerOptions": {
+    "strictNullChecks": false,
```
djones:
```suggestion
```

test/tsconfig.json
```diff
@@ -1,6 +1,8 @@
...
+    "suppressImplicitAnyIndexErrors": true,
```
djones:
```suggestion
```

test/tsconfig.json
djones: Both of these can be pretty easily resolved. I actually find strict null checks to actually be very helpful.

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+      })
+
+      it('adds a guid data-attribute to the original DOM', () => {
+        expect(document.querySelectorAll('[data-percy-element-id]').length).to.equal(5)
+      })
+
+      it('adds matching guids to the orignal DOM and cloned DOM', () => {
+        const originalElementIds = document.querySelectorAll('[data-percy-element-id]')
+        const cloneElementIds = dom.clonedDOM.querySelectorAll('[data-percy-element-id]')
+        const originalValue = originalElementIds[0].attributes['data-percy-element-id'].value
```
djones:
```suggestion
        const originalValue = originalElementIds[0].getAttribute('data-percy-element-id')
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+        const cloneValue = cloneElementIds[0].attributes['data-percy-element-id'].value
```
djones:
```suggestion
        const cloneValue = cloneElementIds[0].getAttribute('data-percy-element-id')
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+  testContainer.innerHTML = finalDOM
+  document.body.appendChild(testContainer)
+
+  return document
+}
+
+// create a stylesheet in the DOM and add rules using the CSSOM
+function createCSSOM() {
+  const style = document.createElement('style')
+  const testingContainer = document.querySelector('.test-container')
```
djones:
```suggestion
  const testingContainer = document.querySelector('.test-container') as HTMLElement
```

this resolves the need for `strictNullChecks` to be disabled.

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
+import { expect } from 'chai'
+import * as cheerio from 'cheerio'
+// @ts-ignore
+import { check, type } from 'interactor.js'
+import * as sinon from 'sinon'
+import DOM from '../../src/percy-agent-client/dom'
+
+// Create valid DOM to pass to the DOM class.
+function createExample(dom: any) {
```
djones:
```suggestion
function createExample(dom: string) {
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+
+// create a stylesheet in the DOM and add rules using the CSSOM
+function createCSSOM() {
+  const style = document.createElement('style')
+  const testingContainer = document.querySelector('.test-container')
+
+  style.type = 'text/css'
+  testingContainer.appendChild(style)
+
+  const cssomStyleSheet = document.styleSheets[0] as any
```
djones:
```suggestion
  const cssomStyleSheet = document.styleSheets[0] as CSSStyleSheet
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    describe('CSSOM with JS enabled', () => {
+      beforeEach(() => {
+        const exampleDOM = createExample('<div class="box"></div>')
+
+        createCSSOM()
+        dom = new DOM(exampleDOM, { enableJavaScript: true })
+      })
+
+      it('does not serialize the CSSOM when JS is enabled', () => {
+        const cssomOwnerNode = document.styleSheets[0].ownerNode as any
```
djones:
```suggestion
        const cssomOwnerNode = document.styleSheets[0].ownerNode as HTMLElement
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    describe('CSSOM with JS disabled', () => {
+      beforeEach(() => {
+        const exampleDOM = createExample('<div class="box"></div>')
+
+        createCSSOM()
+        dom = new DOM(exampleDOM)
+      })
+
+      it('does not mutate the orignal DOM', () => {
+        const cssomOwnerNode = document.styleSheets[0].ownerNode as any
```
djones:
```suggestion
        const cssomOwnerNode = document.styleSheets[0].ownerNode as HTMLElement
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+      it('does not mutate values in origial DOM', () => {
+        const originalDOMInput: any = document.querySelector('#name')
+
+        expect(originalDOMInput.attributes.value).to.equal(undefined)
+        expect(dom.clonedDOM.querySelector('#name').attributes.value.value).to.equal('Bob Boberson')
+      })
+    })
+
+    describe('CSSOM with inputs', () => {
+      let $domString: any
```
djones:
```suggestion
      let $domString: CheerioStatic
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+      it('serializes into the DOM clone', () => {
+        const serializedCSSOM = dom.clonedDOM.querySelectorAll('[data-percy-cssom-serialized]')
+
+        expect(serializedCSSOM.length).to.equal(1)
+        expect(serializedCSSOM[0].innerText).to.equal('.box { height: 500px; width: 500px; background-color: green; }')
+      })
+
+      describe('adding new styles after snapshotting', () => {
+        it('does not break the CSSOM', () => {
+          const cssomSheet = document.styleSheets[0] as any
```
djones:
```suggestion
          const cssomSheet = document.styleSheets[0] as CSSStyleSheet
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+      it('does not serialize the CSSOM when JS is enabled', () => {
+        const cssomOwnerNode = document.styleSheets[0].ownerNode as any
+
+        expect(cssomOwnerNode.innerText).to.equal('')
+        expect(dom.clonedDOM.querySelectorAll('[data-percy-cssom-serialized]').length).to.equal(0)
+      })
+    })
+
+    describe('inputs', () => {
+      let $domString: any
```
djones:
```suggestion
      let $domString: CheerioStatic
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+        const originalElementIds = document.querySelectorAll('[data-percy-element-id]')
+        const cloneElementIds = dom.clonedDOM.querySelectorAll('[data-percy-element-id]')
+        const originalValue = originalElementIds[0].attributes['data-percy-element-id'].value
+        const cloneValue = cloneElementIds[0].attributes['data-percy-element-id'].value
+
+        expect(originalValue).to.equal(cloneValue)
+      })
+
+      it('does not mutate values in origial DOM', () => {
+        const originalDOMInput: any = document.querySelector('#name')
```
djones:
```suggestion
        const originalDOMInput = document.querySelector('#name') as HTMLElement
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+
+        expect(originalDOMInput.attributes.value).to.equal(undefined)
```
djones:
```suggestion
        expect(originalDOMInput.getAttribute('value')).to.equal(undefined)
```

I think your intention is something like this?

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+        expect(dom.clonedDOM.querySelector('#name').attributes.value.value).to.equal('Bob Boberson')
```
djones:
```suggestion
        expect(dom.clonedDOM.querySelector('#name').getAttribute('value')).to.equal('Bob Boberson')
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+  })
+
+  describe('passing a DOM transform option', () => {
+    let consoleStub: any
+
+    beforeEach(() => {
+      consoleStub = sinon.stub(console, 'error')
+      dom = new DOM(createExample('<span class="delete-me">Delete me</span>'), {
+        domTransformation(dom: any) {
+          const span = dom.querySelector('.delete-me')
```
djones:
```suggestion
          const span = dom.querySelector('.delete-me') as HTMLSpanElement
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    })
+  })
+
+  describe('passing a DOM transform option', () => {
+    let consoleStub: any
+
+    beforeEach(() => {
+      consoleStub = sinon.stub(console, 'error')
+      dom = new DOM(createExample('<span class="delete-me">Delete me</span>'), {
+        domTransformation(dom: any) {
```
djones:
```suggestion
        domTransformation(dom) {
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+      dom = new DOM(document)
+    })
+
+    it('always has a doctype', () => {
+      expect(dom.snapshotString()).to.contain('<!DOCTYPE html>')
+    })
+  })
+
+  describe('passing a DOM transform option', () => {
+    let consoleStub: any
```
djones:
```suggestion
    let consoleStub: sinon.SinonStub
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+
+      it('does not mutate the orignal DOM', () => {
+        const cssomOwnerNode = document.styleSheets[0].ownerNode as any
+
+        expect(cssomOwnerNode.innerText).to.equal('')
+        expect(document.querySelectorAll('[data-percy-cssom-serialized]').length).to.equal(0)
+      })
+
+      it('serializes into the DOM clone', () => {
+        const serializedCSSOM = dom.clonedDOM.querySelectorAll('[data-percy-cssom-serialized]')
```
djones:
```suggestion
        const serializedCSSOM = Array.from(
          dom.clonedDOM.querySelectorAll('[data-percy-cssom-serialized]'),
        ) as HTMLElement[]
```

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+        $domString = cheerio.load(dom.snapshotString())
+      })
+
+      it('serializes the input value', () => {
+        expect($domString('#name').attr('value')).to.equal('Bob Boberson')
+        expect(dom.clonedDOM.querySelector('#name').attributes.value.value).to.equal('Bob Boberson')
+      })
+
+      it('serializes the CSSOM', () => {
+        const serializedCSSOM = dom.clonedDOM.querySelectorAll('[data-percy-cssom-serialized]')
```
djones:
```suggestion
        const serializedCSSOM = Array.from(
          dom.clonedDOM.querySelectorAll('[data-percy-cssom-serialized]'),
        ) as HTMLElement[]
```

src/percy-agent-client/dom.ts
Robdel12: Yeah that's a real good one 👍

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+   * Serialize parts of the DOM that aren't preserved in a deep clone. Anything
+   * that is not encoded in the DOM tree needs to be accounted for here (CSSOM,
+   * input values, canvas, etc).
+   *
+   * This method should always capture these values from the original DOM and
+   * serialize into the cloned DOM. Never mutate the original DOM.
+   *
+   */
+  private stabilizeDOM(clonedDOM: HTMLDocument): HTMLElement {
+    let stabilizedDOMClone = this.serializeInputElements(clonedDOM)
```
Robdel12: Just a thought here.. We don't need to pass this var around since everything that's done here mutates the cloned DOM.

src/percy-agent-client/dom.ts
```diff
@@ -0,0 +1,213 @@
...
+    let dom = this.clonedDOM as any
+    const doctype = this.getDoctype()
+
+    // Sometimes you'll want to transform the DOM provided into one ready for snapshotting
+    // For example, if your test suite runs tests in an element inside a page that
+    // lists all yours tests. You'll want to "hoist" the contents of the testing container to be
+    // the full page. Using a dom transformation is how you'd achieve that.
+    if (this.options.domTransformation) {
+      try {
+        dom = this.options.domTransformation(dom)
```
Robdel12: Should we hand them a clone of the clone? Anything they do here mutative. 🤔

test/percy-agent-client/dom.test.ts
```diff
@@ -0,0 +1,260 @@
...
+        createCSSOM()
+        await type('#name', 'Bob Boberson')
+
+        dom = new DOM(example)
+        $domString = cheerio.load(dom.snapshotString())
+      })
+
+      it('serializes the input value', () => {
+        expect($domString('#name').attr('value')).to.equal('Bob Boberson')
+        expect(dom.clonedDOM.querySelector('#name').attributes.value.value).to.equal('Bob Boberson')
```
djones:
```suggestion
        expect(dom.clonedDOM.querySelector('#name').getAttribute('value')).to.equal('Bob Boberson')
```

PR #217, opened 2019-05-30
fix: [Security] Upgrade axios

Robdel12:
## What is this?

There's a major security issue in older versions of axios, which is what we use for network requests in Agent. This will close this: https://github.com/percy/percy-agent/network/alert/package-lock.json/axios/open

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6208,8 +6208,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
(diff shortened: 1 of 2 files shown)


Robdel12:
Ha, this breaks something.. Can't reproduce locally though.

Robdel12:
Keeping an eye on https://github.com/axios/axios/issues/2183 for 0.18.1 👀

PR #227, opened 2019-06-04
fix: Always apply form elements value property as an attribute

Robdel12:
## What is this?

Previously, if the `value` attribute was present in the DOM, we didn't serialize the elements `value` _property_ to the attribute. There's no reason for us to make this check. it's 100% safe for us to always apply the property value as an attribute on the DOM node. 

I also noticed there's a bug with the switch where we didn't have a `break` on `textarea`s, so we were always adding an invalid `value` attribute.

### DIFF
--- a/src/percy-agent-client/dom.ts
+++ b/src/percy-agent-client/dom.ts
@@ -121,19 +121,16 @@ class DOM {
       switch (elem.type) {
         case 'checkbox':
         case 'radio':
-          if (elem.checked && !elem.hasAttribute('checked')) {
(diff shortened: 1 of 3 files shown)


djones:
@blake-newman would you be able to confirm this fixes your use case?

blake-newman:
@djones yes I can no problem, can't right now as our ci infrastructure turns off during OOO hours. So will have report back to you for your morning

blake-newman:
Can confirm this works :+1:

PR #243, opened 2019-06-06
feat: Improved large resource handling

djones:
This PR:

* Skips resources and snapshots early on that are destined to fail on upload anyway
* Reduces noise in debug logs with large dom snapshots 
* Debug logs when and why we're skipping a given resource or snapshot

You'll see logs like this in the integration test run:

```
[percy] snapshot skipped[max_file_size_exceeded]: 'Test case: exceeds-dom-snapshot-size-limit.html'
```

That is excepted behavior. Graceful handling of a large DOM and warning you about it.

This PR also drops large resources gracefully and early. If you have `LOG_LEVEL=debug` set you will see in the logs something like this:

```
[percy] Skipping [max_file_size_exceeded_21141605] [1280 px]: http://localhost:8000/20mb.png
```

### DIFF
--- a/src/services/agent-service.ts
+++ b/src/services/agent-service.ts
@@ -69,15 +69,20 @@ export default class AgentService {
   private async handleSnapshot(request: express.Request, response: express.Response) {
     profile('agentService.handleSnapshot')
 
-    // truncate domSnapshot becuase it can be very noisy in the logs
-    if (request.body.domSnapshot.length > Constants.MAX_LOG_LENGTH) {
-      request.body.domSnapshot = request.body.domSnapshot.substring(0, Constants.MAX_LOG_LENGTH)
-      request.body.domSnapshot += `[truncated at ${Constants.MAX_LOG_LENGTH}]`
+    // truncate domSnapshot for the logs if it's very large
+    let domSnapshotLog = request.body.domSnapshot
+    if (domSnapshotLog.length > Constants.MAX_LOG_LENGTH) {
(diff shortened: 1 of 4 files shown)


src/services/agent-service.ts
```diff
@@ -68,9 +68,20 @@ export default class AgentService {
   private async handleSnapshot(request: express.Request, response: express.Response) {
     profile('agentService.handleSnapshot')
 
+    // truncate domSnapshot for the logs if it's very large
+    let domSnapshotLog = request.body.domSnapshot
+    if (domSnapshotLog.length > Constants.MAX_LOG_LENGTH) {
+      domSnapshotLog = domSnapshotLog.substring(0, Constants.MAX_LOG_LENGTH)
```
djones: I think you'll like this one @Robdel12. We'll no longer spit out huge logs that are 99% the `domSnapshot` content.

src/services/response-service.ts
```diff
@@ -55,6 +56,13 @@ export default class ResponseService extends PercyClientService {
       return
     }
 
+    const responseBodySize = (await response.buffer()).length
+    if (responseBodySize > Constants.MAX_FILE_SIZE_BYTES) {
```
fotinakis: Per discussion, let's reorganize this to `stat` the local file copy afterward, instead of reading the buffer and checking it here before writing the local copy (ie. optimize for the 99% case of not skipping, and use stat because it's fast and avoids the double buffer read)

djones: Fixed with https://github.com/percy/percy-agent/pull/243/commits/941db82a027f68cf0078319e211386e4d9f34936

PR #257, opened 2019-06-18
feat: Expanded `.percy.yml` configuration support

djones:
## New configuration options in `.percy.yml`

On the surface this PR gives us the ability to now configure these options in the `.percy.yml` file:

* static-snapshots
  * path
  * port (_the port your static site ends up hosted on_)
* agent
  * port (_configurable but not fully functional yet though_)
  * asset-discovery
    * network-idle-timeout
    * page-pool-size-min
    * page-pool-size-max

## Configuration handling refactor

Deeper into the PR, you'll find a total refactor of the way we handle configuration in @percy/agent. As we've expanded what you can configure in `.percy.yml`, it's become more important to be able to layer config from various places on top of each other.

These configuration layers are setup like this:

1) default configuration
2) `.percy.yml` configuration
3) supplied command line flags
4) supplied command line arguments

The `ConfigurationService` can handle input from all of these places and produce one unified configuration for use by other service objects.

## Page pool size experiment 

Now that you can configure `page-pool-size` I ran some experiments to understand the resource differences. The test I ran took 101 snapshots with 3 widths on a very simple todo app page and it sent them up in parallel. Here's the script `snapshots.js` I used:

```javascript
const PercyScript = require('@percy/script');

PercyScript.run(async (page, percySnapshot) => {
  let i = 0
  await page.goto('http://localhost:8000');
  const snapshots = []
  const parallelTotal = 100

  while (i <= parallelTotal) {
     snapshots.push(percySnapshot(`Parallel Snapshot ${i}`))
     i++
  }

  await Promise.all(snapshots)
});
```

and here's the results with different page-pool settings. I ran `time npm run snapshots`.

```bash
# page-pool-size-min: 1
# page-pool-size-max: 1
# network-idle-timeout: 50
npm run snapshots  23.56s user 5.14s system 78% cpu 36.328 total
npm run snapshots  24.01s user 5.21s system 78% cpu 37.019 total
npm run snapshots  22.22s user 4.90s system 76% cpu 35.666 total

# page-pool-size-min: 1
# page-pool-size-max: 3
# network-idle-timeout: 50
npm run snapshots  20.16s user 4.62s system 141% cpu 17.497 total
npm run snapshots  21.74s user 4.84s system 147% cpu 18.058 total
npm run snapshots  21.61s user 4.84s system 147% cpu 17.905 total

# page-pool-size-min: 1
# page-pool-size-max: 5
# network-idle-timeout: 50
npm run snapshots  25.29s user 5.32s system 185% cpu 16.503 total
npm run snapshots  28.69s user 5.98s system 210% cpu 16.508 total
npm run snapshots  24.54s user 5.35s system 180% cpu 16.519 total

# page-pool-size-min: 1
# page-pool-size-max: 10
# network-idle-timeout: 50
npm run snapshots  25.84s user 5.44s system 199% cpu 15.691 total
npm run snapshots  28.79s user 5.95s system 209% cpu 16.597 total
npm run snapshots  28.06s user 5.82s system 198% cpu 17.105 total

# page-pool-size-min: 3
# page-pool-size-max: 5
# network-idle-timeout: 50
npm run snapshots  24.88s user 5.27s system 181% cpu 16.575 total
npm run snapshots  25.17s user 5.22s system 185% cpu 16.371 total
npm run snapshots  25.70s user 5.35s system 190% cpu 16.296 total

# page-pool-size-min: 3
# page-pool-size-max: 5
# network-idle-timeout: 0 <- not recommended
npm run snapshots  24.84s user 5.16s system 185% cpu 16.194 total
npm run snapshots  25.12s user 5.27s system 191% cpu 15.897 total
npm run snapshots  33.71s user 6.57s system 246% cpu 16.362 total
````

For each page open in the pool, I noticed about ~130mb of RAM was used. 

What I observe from this is the following:

* If you would like a good tradeoff of speed and resources just leave these settings as the default (min 1, max 5).
* If you want `@percy/agent` to take up the minimum amount of resources (CPU and memory) and aren't concerned about speed, you should set your min and max pool size to 1.
* If you have a large machine running your tests you might find a small performance boost by setting your max pool size to something larger, like 10 or 20.
* Setting a min size to something higher than 1 didn't really yield any benefit. `@percy/agent` will quickly scale up to the max pool size if it's needed.

### CODE REVIEW

File: test/support/.percy.yml [hunk 1]
```diff
@@ -1,8 +1,16 @@
...
-  min-height: 1024
+  min-height: 1024 # px
 static-snapshots:
+  path: _site/
+  port: 9999
   base-url: /blog/
   snapshot-files: '**/*.html'
   ignore-files: '**/*.htm'
+agent:
+  port: 1111
```
djones: this is questionable about if this should be here since it's not really supported properly.

src/services/agent-service-constants.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_PORT: number = 5338 // BEES
```
djones: these had to be explicitly split out because they are imported into `PercyAgentClient`.

src/services/static-snapshot-service.ts
```diff
@@ -73,19 +74,27 @@ export default class StaticSnapshotService {
...
-    const baseUrl = this._buildLocalUrl()
-    const pageUrls = [] as any
+    // We very intentially remove '' values from these globs because that matches every file
+    const ignoreGlobs = this.configuration['ignore-files']
+      .split(',')
+      .filter((value) => value !== '')
+
+    const snapshotGlobs = this.configuration['snapshot-files']
+      .split(',')
+      .filter((value) => value !== '')
```
djones: previously we were not guarding against `''` values for the snapshot globs, which appear to match everything. This is a bit of a bonus change.

src/services/asset-discovery-service.ts
```diff
@@ -1,46 +1,38 @@
...
   }
 
   async setup() {
     profile('-> assetDiscoveryService.setup')
+
     const browser = this.browser = await this.createBrowser()
     this.pagePool = await this.createPagePool(() => {
       return this.createPage(browser)
-    }, this.PAGE_POOL_SIZE_MIN, this.PAGE_POOL_SIZE_MAX)
+    }, this.configuration['page-pool-size-min'],
```
djones: this is one of the key bits of this PR. We're now supplying a configurable min/max pool size.

package.json
```diff
@@ -22,6 +22,7 @@
     "@oclif/config": "^1",
     "@oclif/plugin-help": "^2",
     "@oclif/plugin-not-found": "^1.2",
+    "@types/deepmerge": "^2.2.0",
```
Robdel12: I've never understood this -- do these need to be dependencies of our package? Since agent is distributed as a compiled package it feels weird that everyone who installs our package also gets a bunch of types packages. too.

src/services/agent-service.ts
```diff
@@ -30,20 +30,24 @@ export default class AgentService {
...
   }
 
-  async start(options: AgentOptions = {}) {
+  async start(configuration: Configuration) {
     this.buildId = await this.buildService.create()
 
     if (this.buildId !== null) {
-      this.server = this.app.listen(options.port)
-      this.snapshotService = new SnapshotService(this.buildId, {networkIdleTimeout: options.networkIdleTimeout})
+      console.log(`DAVE: ${configuration.agent.port}`)
```
Robdel12: DAVE 😆

test/support/.percy.yml
Robdel12: yeah, I think we should create a follow up ticket to fix it. I've seen a few folks in support try to use it

package.json
Robdel12: This is a good SO for this: https://stackoverflow.com/a/46011417

src/services/agent-service.ts
djones:
```
\ /
 |
/o\
```

package.json
djones: Shifted all the types to dev deps in 65f535220ae7d0dc4b299524026b862048c02333

PR #270, opened 2019-07-02
fix: Static service Wait for network idle

Robdel12:
## What is this? 

I've been seeing a lot of errors like:

```
Error: Execution context was destroyed, most likely because of a navigation.
```

This happens because we try to start altering the page (by injecting our JS) before everything is settled down. This change won't promise the page has actually loaded (that's very hard to determine) but it will wait for most network requests to resolve before trying do anything.


djones:
If this still brings up troubles, we could use https://github.com/percy/percy-agent/blob/master/src/utils/wait-for-network-idle.ts next

Robdel12:
@djones OHH very nice, that would be a great next step

PR #273, opened 2019-07-03
fix: Reject redirected requests for static snapshots

Robdel12:
## What is this?

This makes it so we no longer get redirected from the URL we visit to another one. That usually results in an error, which fails the test suite.

This also adds some try/catches around parts of the snapshot that could fail with helpful error messages

### DIFF
--- a/src/services/static-snapshot-service.ts
+++ b/src/services/static-snapshot-service.ts
@@ -44,27 +44,46 @@ export default class StaticSnapshotService {
     const percyAgentClientFilename = agentJsFilename()
     const page = await browser.newPage()
 
+    // Do not follow redirects to ensure we don't navigate to another page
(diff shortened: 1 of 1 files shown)


src/services/static-snapshot-service.ts
```diff
@@ -44,27 +44,46 @@ export default class StaticSnapshotService {
...
-
-      await page.evaluate((url) => {
-        const percyAgentClient = new PercyAgent()
-        const parsedURL = new URL(url)
-        const snapshotName = parsedURL.pathname || url
-
-        return percyAgentClient.snapshot(snapshotName)
-      }, url)
+      try {
+        await page.goto(url, { waitUntil: 'networkidle2' })
```
Robdel12: `networkidle2` is much faster than `networkidle0`

PR #295, opened 2019-07-15
feat: Serialize select elements values

Robdel12:
## What is this?

This will capture the selected values from `<select>` elements. 

### TODOs:

- [x] Fight TypeScript to maybe get union types going..
- [x] Remove `scoped` import from `dom.test.ts` file

### DIFF
--- a/src/percy-agent-client/dom.ts
+++ b/src/percy-agent-client/dom.ts
@@ -3,6 +3,8 @@ export interface DOMOptions {
   domTransformation?: (dom: HTMLDocument) => void
 }
 
+const FORM_ELEMENTS_SELECTOR = 'input, textarea, select'
+
 /**
  * A single class to encapsulate all DOM operations that need to be performed to
  * capture the customer's application state.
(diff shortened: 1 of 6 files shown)


src/percy-agent-client/dom.ts
```diff
@@ -110,20 +112,39 @@ class DOM {
...
-    const formNodes = this.originalDOM.querySelectorAll('input, textarea')
+    const formNodes = this.originalDOM.querySelectorAll(FORM_ELEMENTS_SELECTOR)
     const formElements = Array.from(formNodes) as HTMLFormElement[]
 
     formElements.forEach((elem) => {
       const inputId = elem.getAttribute('data-percy-element-id')
       const selector = `[data-percy-element-id="${inputId}"]`
-      const cloneEl = clonedDOM.querySelector(selector) as HTMLInputElement
+      // TODO: lol, TypeScript. This seems bad... They're not all selects and TS is totally fine with it.
+      const cloneEl = clonedDOM.querySelector(selector) as HTMLSelectElement
```
Robdel12: @djones it's either this or make it any 👀

src/percy-agent-client/dom.ts
```diff
@@ -110,20 +112,38 @@ class DOM {
...
   private serializeInputElements(clonedDOM: HTMLDocument): void {
-    const formNodes = this.originalDOM.querySelectorAll('input, textarea')
+    const formNodes = this.originalDOM.querySelectorAll(FORM_ELEMENTS_SELECTOR)
     const formElements = Array.from(formNodes) as HTMLFormElement[]
 
     formElements.forEach((elem) => {
       const inputId = elem.getAttribute('data-percy-element-id')
       const selector = `[data-percy-element-id="${inputId}"]`
-      const cloneEl = clonedDOM.querySelector(selector) as HTMLInputElement
+      const cloneEl = clonedDOM.querySelector(selector) as any
```
Robdel12: I called it out in the commit that made this change, but want to be clear here too.

I made this any since we're looping over an array of DOM nodes that can be any of these (at this time): `select`, `input`, `textarea`

Since TS doesn't allow you to recast a type, I'd rather make this `any` since the type of node can/will change with each loop.

PR #317, opened 2019-08-06
fix: Use `innerHTML` rather than `innerText` for textareas

Robdel12:
## What is this?

This is so we can properly capture HTML content in textareas (which is common with WYSIWYG editors) 

### Question

Will this have any impact on non-HTML content? Or will `innerHTML` work as expected?

### DIFF
--- a/src/percy-agent-client/dom.ts
+++ b/src/percy-agent-client/dom.ts
@@ -146,8 +146,8 @@ class DOM {
 
           break
         case 'textarea':
-          // setting text or value does not work but innerText does
-          cloneEl!.innerText = elem.value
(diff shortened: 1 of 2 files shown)


src/percy-agent-client/dom.ts
```diff
@@ -147,7 +147,7 @@ class DOM {
           break
         case 'textarea':
           // setting text or value does not work but innerText does
```
wwilsman: We should update this comment too

Robdel12: Heh, I realized that and then amended the commit

PR #318, opened 2019-08-06
fix: Log error if there is one from the config service

Robdel12:
## What is this?

When users are having issues with their config file in support it would be helpful to log the error that's being caught in the config service. I also added another debug log that logs out the directory we'd be expecting the `.percy.yml` file to be.

### DIFF
--- a/src/services/configuration-service.ts
+++ b/src/services/configuration-service.ts
@@ -22,12 +22,13 @@ export default class ConfigurationService {
     try {
       const userConfigFilePath = path.join(process.cwd(), configurationFile)
       const userConf = yaml.safeLoad(fs.readFileSync(userConfigFilePath, 'utf8'))
+      logger.debug(`Current config file path: ${userConfigFilePath}`)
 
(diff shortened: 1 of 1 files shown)


src/services/configuration-service.ts
```diff
@@ -22,12 +22,13 @@ export default class ConfigurationService {
...
       const userConf = yaml.safeLoad(fs.readFileSync(userConfigFilePath, 'utf8'))
+      logger.debug(`Current config file path: ${userConfigFilePath}`)
 
       // apply a deep overwrite merge to userConf and this.configuration
       const overwriteMerge = (destinationArray: any, sourceArray: any, options: any) => sourceArray
       this.configuration = deepmerge(this.configuration, userConf, { arrayMerge: overwriteMerge })
-    } catch {
-      logger.debug('.percy.yml configuration file not supplied or failed to be loaded and parsed.')
+    } catch (error) {
+      logger.debug(`.percy.yml configuration file not supplied or failed to be loaded and parsed: ${error}`)
```
djones: :+1:

PR #320, opened 2019-08-07
feat: Support capturing assets from allowed hostnames

timhaines:
Adds support for allowed hostnames that assets should be captured from.

If a hostname is specified, any requests to assets on that hostname will be captured, and they'll be uploaded to Percy's API as snapshot resources.  During rendering time, **http** requests to the captured asset's url will be hijacked, and the captured resource served. 

Hostnames can be provided either with the `-allowed-hostname` or `-h` flag, or they can be specified in the `.percy.yml` file as an array under agent -> asset-discovery -> allowed-hostnames.

The main function is tested with the new alternate-hostname integration test. It loads a vue.svg file from a hostname of localhost.me. Localhost.me resolves to 127.0.0.1, so we can serve it from our local dev/CI environment, and be sure that the captured asset is being served in production. 

Note: requests to https assets won't be successfully highjacked in our rendering environment yet, but this PR should include the appropriate support for capturing them to our @percy/agent based SDKs.

### DIFF
--- a/test/integration/agent-integration.test.ts
+++ b/test/integration/agent-integration.test.ts
@@ -1,5 +1,4 @@
 import * as cheerio from 'cheerio'
-import * as fs from 'fs'
 import { Server } from 'http'
 import * as httpServer from 'http-server'
 import { describe } from 'mocha'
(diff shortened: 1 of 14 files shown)


test/integration/agent-integration.test.ts
```diff
@@ -1,5 +1,4 @@
 import * as cheerio from 'cheerio'
-import * as fs from 'fs'
```
Robdel12: 👏

test/integration/testcases/alternate-hostname.html
```diff
@@ -0,0 +1,18 @@
...
+  <div>
+    <p>
+      If alternate hostname support is working, there should be an svg image of the vue logo below.
+    </p>
+    <p>
+      It works because localhost.me resolves to 127.0.0.1, allowing us to serve the vue logo over
+      it, and we start the integration tests by specifying localhost.me as an allowed-hostname
+      with the -h flag.
+    </p>
+    <img src="http://localtest.me:8000/vue.svg">
```
Robdel12: Curious how this works in CI? Are we editing a host file to make this work?

timhaines: @Robdel12 no, `localtest.me`'s dns entry points to 127.0.0.1.  So in CI the node server we spin up that listens to 127.0.0.01:8000 works just the same way it does locally, with no configuration needed.

PR #332, opened 2019-08-22
fix: Stop and finalize build on exec termination

wwilsman:
## Purpose

When running `percy exec` with a command that causes the process to hang, terminating the process causes builds to get stuck in a receiving state due to the build never finalizing.

## Approach

Add [signal handlers](https://nodejs.org/api/process.html#process_signal_events) for `SIGHUP`, `SIGINT`, and `SIGTERM` to stop and finalize the build.

Multiple signals are sometimes sent and cause the stop method to get called multiple times. This didn't seem to effect finalizing builds, but caused noisy output in the terminal. An instance property, `exiting`, is used to bail out of the stop method when it has already been called. This also prevents secondary events from exiting before the first event has finalized the build.

### Other changes

I also corrected a check for `percyEnabled` which treated the property as a getter when it is actually a method. Fixing this also highlighted a test which actually tested the incorrect behavior. This test was updated to reflect the desired behavior, which is to not warn about a missing token when Percy is disabled.

### DIFF
--- a/src/commands/exec.ts
+++ b/src/commands/exec.ts
@@ -1,4 +1,4 @@
-import {flags} from '@oclif/command'
+import { flags } from '@oclif/command'
 import * as spawn from 'cross-spawn'
 import { DEFAULT_CONFIGURATION } from '../configuration/configuration'
 import ConfigurationService from '../services/configuration-service'
@@ -35,9 +35,7 @@ export default class Exec extends PercyCommand {
   async run() {
     await super.run()
 
-    const {argv} = this.parse(Exec)
-    const {flags} = this.parse(Exec)
-
+    const { argv, flags } = this.parse(Exec)
     const command = argv.shift()
 
     if (!command) {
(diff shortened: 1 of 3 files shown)


src/commands/percy-command.ts
```diff
@@ -22,7 +22,7 @@ export default class PercyCommand extends Command {
   }
 
   async run() {
-    if (this.percyEnabled && !this.percyTokenPresent()) {
+    if (this.percyEnabled() && !this.percyTokenPresent()) {
```
Robdel12: Great catch here 👍

src/commands/exec.ts
```diff
@@ -54,14 +55,23 @@ export default class Exec extends PercyCommand {
     }
 
     // Even if Percy will not run, continue to run the subprocess
-    const spawnedProcess = spawn(command, argv, {stdio: 'inherit'})
+    const spawnedProcess = spawn(command, argv, { stdio: 'inherit' })
+    spawnedProcess.on('exit', (code) => this.stop(code))
+
+    // Recieving any of these events should stop the agent and exit
```
Robdel12: Tiny typo `Receiving`

src/commands/percy-command.ts
djones: Nice catch!

src/commands/exec.ts
wwilsman: `i` before `e`, except _in all of these other cases_ (╯°□°）╯︵ ┻━┻

Robdel12: ┬─┬ノ(ಠ_ಠノ)

Just wanna say I caught a misspelling 👀

PR #341, opened 2019-09-03
feat: Add configuration option to enable javascript

wwilsman:
## Purpose

Enabling JavaScript must be done on a per-snapshot basis and in rare cases where the user needs to enable it for every snapshot, it becomes tedious to add it to every snapshot call.

## Approach

Add an `enable-javascript` option to the configuration file to enable setting this option for all snapshots in a single setting.

I know we don't recommend setting this option _normally_, however if needed it can be a pain to set this for every single snapshot.

### DIFF
--- a/src/services/agent-service.ts
+++ b/src/services/agent-service.ts
@@ -92,7 +92,9 @@ export class AgentService {
     const configuration = new ConfigurationService().configuration
     const snapshotOptions: SnapshotOptions = {
       widths: request.body.widths || configuration.snapshot.widths,
-      enableJavaScript: request.body.enableJavaScript,
+      enableJavaScript: request.body.enableJavaScript != null
+        ? request.body.enableJavaScript
+        : configuration.snapshot['enable-javascript'],
       minHeight: request.body.minHeight || configuration.snapshot['min-height'],
     }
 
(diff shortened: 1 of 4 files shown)


src/services/agent-service.ts
```diff
@@ -92,7 +92,9 @@ export class AgentService {
     const configuration = new ConfigurationService().configuration
     const snapshotOptions: SnapshotOptions = {
       widths: request.body.widths || configuration.snapshot.widths,
-      enableJavaScript: request.body.enableJavaScript,
+      enableJavaScript: request.body.enableJavaScript != null
+        ? request.body.enableJavaScript
+        : configuration.snapshot['enable-javascript'],
```
wwilsman: The [nullish operator](https://github.com/microsoft/TypeScript/issues/26578) would've been perfect here:

```js
request.body.enableJavaScript ?? configuration.snapshot['enable-javascript']
```

Robdel12:
Oh, also don't forget to update the SDK config docs

wwilsman:
I don't think the docs need an update since this is a "hidden" option. We don't document enableJavaScript anywhere else, do we? If we wanted to, I'd say we should update all docs across the board

PR #346, opened 2019-09-16
feat: Introduce an API for Percy Specific CSS

Robdel12:
## What is this?

This PR introduces first class support for providing CSS specific to only Percy's rendering environment though the SDKs. Now users will be able to provide a new option called `percyCSS` with our SDKs. This can be set globally and on a per-snapshot basis. 

## Approach

Initially the implementation of this was going to be fairly easy. We would take a new snapshot option and then serialize that CSS into the DOM when we're serializing inputs and what not. It turns out, CSP can block this injected CSS (thank you SDK test site!). 

To work around CSP issues, we instead now take that CSS and create a new resource for this snapshot which contains the CSS. If you have global CSS, we'll concat that with the per-snapshot CSS. This file is named with the current time stamp + `percy-specific.css`. It does not use the `@percy` media query since this CSS can only ever exist in Percy)

This CSS will be applied at the very bottom of their DOM snapshot, before the closing `</body>` tag. This hopefully will ensure the cascade will do its job and `!important` won't be needed often. 

## Usage

Now there's a simple API for providing Percy only CSS. For snapshot specific css:

```js
await percySnapshot('Home page', { percyCSS: `iframe { display: none; }` })
```

For global CSS in the `.percy.yml` file:

```yml
version: 1
snapshot:
  widths: [375, 1280]
  percy-css: | 
    iframe { 
      display none; 
    }
```

## TODOs once this ships

- [ ] Update SDK config docs
- [ ] Update Percy specific CSS docs
- [ ] Ensure all SDKs are passing `options` correctly.


.circleci/config.yml
```diff
@@ -36,6 +36,9 @@ jobs:
       - run:
           name: Client tests
           command: $NYC yarn test-client --singleRun
+      - run:
+          name: Setup integration test .percy.yml
+          command: mv .ci.percy.yml .percy.yml
```
Robdel12: I'm not crazy about this, but it's needed because adding a `.percy.yml` file breaks the unit tests. 

This can go away once you can pass a custom file path for the `.yml` file. Then the unit tests can have their own mocked & static .yml file.

wwilsman:
Haven't reviewed yet, but I just wanted to add that the `yml` option might be nice to document like this so it's clear how multiple rules might look:

```yml
version: 1
snapshot:
  widths: [375, 1280]
  percy-css: |
    iframe {
      display none;
    }
```

src/services/agent-service.ts
```diff
@@ -94,15 +94,25 @@ export class AgentService {
     if (!this.snapshotService) { return response.json({success: false}) }
 
     const configuration = new ConfigurationService().configuration
+    // trim the string of whitespace and concat per-snapshot CSS with the globally specified CSS
+    const percySpecificCSS = configuration.snapshot['percy-css'].concat(request.body.percyCSS || '').trim() || ''
```
wwilsman: The final `|| ''` is redundant since `.trim()` will fail if it's not a string and will otherwise return an empty string.

src/services/agent-service.ts
```diff
@@ -94,15 +94,25 @@ export class AgentService {
...
       widths: request.body.widths || configuration.snapshot.widths,
       enableJavaScript: request.body.enableJavaScript != null
         ? request.body.enableJavaScript
         : configuration.snapshot['enable-javascript'],
       minHeight: request.body.minHeight || configuration.snapshot['min-height'],
     }
 
-    const domSnapshot = request.body.domSnapshot
+    let domSnapshot = request.body.domSnapshot
+    const percyCSSFileName = `${Date.now()}-percy-specific.css` as string
```
wwilsman: Pedantic, but if this needs a timestamp I think `percy-specific.${Date.now()}.css` would be easier to find in the filesystem at a glance / search.

src/services/agent-service.ts
```diff
@@ -94,15 +94,25 @@ export class AgentService {
...
+
+    // Inject the link to the percy specific css if the option is passed
+    if (snapshotOptions.percyCSS) {
+      const cssLink = `<link data-percy-specific-css rel="stylesheet" href="/${percyCSSFileName}" />`
+      domSnapshot = domSnapshot.replace(/<\/body>/i, cssLink  + '$&')
```
wwilsman: Apple keyboard strikes again - two spaces before the `+`

src/services/agent-service.ts
Robdel12: Oh, duh, right.

src/services/agent-service.ts
Robdel12: I'm cool with it, matches the log files too

PR #347, opened 2019-09-18
feat: Allow wildcard hostnames

wwilsman:
## Purpose

When the subdomain of a hostname is variable, not known, or has many other subdomains, it is more concise to use a wildcard subdomain.

## Approach

Attempted to use [`domainMatch`](https://www.npmjs.com/package/domain-match) to compare the resource hostname with the allowed hostnames. However internally, the package uses `parsedURL.host` which contains the port number and causes the check to fail. [The source code](https://github.com/palanik/domain-match/blob/master/index.js) was fairly small and straightforward, so I copied it here and adjusted it to use `hostname` which does not contain the port number.

Copying the small util here also made typescript happy about missing type defs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
     "preversion": "yarn clean",
     "test": "yarn build-client && PERCY_TOKEN=abc mocha --forbid-only \"test/**/*.test.ts\" --exclude \"test/percy-agent-client/**/*.test.ts\" --exclude \"test/integration/**/*\"",
     "test-client": "karma start ./test/percy-agent-client/karma.conf.js",
-    "test-integration": "yarn build-client && node ./bin/run exec -h localtest.me -- mocha test/integration/**/*.test.ts",
+    "test-integration": "yarn build-client && node ./bin/run exec -h *.localtest.me -- mocha test/integration/**/*.test.ts",
     "test-snapshot-command": "./bin/run snapshot test/integration/test-static-site -b /dummy-base-url -i '(red-keep)' -c '\\.(html)$'",
     "version": "oclif-dev readme && git add README.md",
     "watch": "npm-watch"
--- a/src/services/response-service.ts
+++ b/src/services/response-service.ts
@@ -4,6 +4,7 @@ import * as os from 'os'
 import * as path from 'path'
 import * as puppeteer from 'puppeteer'
 import { URL } from 'url'
+import domainMatch from '../utils/domain-match'
 import Constants from './constants'
 import PercyClientService from './percy-client-service'
 import ResourceService from './resource-service'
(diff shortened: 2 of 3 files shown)


src/services/response-service.ts
```diff
@@ -1,4 +1,7 @@
 import * as crypto from 'crypto'
+// @ts-ignore - TS hates JS without types
+// why does this default export function need to be imported like this?
```
Robdel12: We can probably remove the comment since the answer is weird stuff that TS does 🤷‍♂ https://stackoverflow.com/a/55055982

wwilsman:
RIP. The asset wasn't captured. Let me see why locally

package.json
```diff
@@ -142,7 +143,7 @@
     "preversion": "yarn clean",
     "test": "yarn build-client && PERCY_TOKEN=abc mocha --forbid-only \"test/**/*.test.ts\" --exclude \"test/percy-agent-client/**/*.test.ts\" --exclude \"test/integration/**/*\"",
     "test-client": "karma start ./test/percy-agent-client/karma.conf.js",
-    "test-integration": "yarn build-client && node ./bin/run exec -h localtest.me -- mocha test/integration/**/*.test.ts",
+    "test-integration": "yarn build-client && node ./bin/run exec -h *.localtest.me -- mocha test/integration/**/*.test.ts",
```
Robdel12: Would this work? It's going to match all subdomains with that host name, but maybe not the hostname itself?

- `images-cdn.localtest.me` would match
- `css-cdn.localtest.me` would match
- `localtest.me` wouldn't match? 🤔

wwilsman: According to [this](https://www.npmjs.com/package/domain-match), it should match. But I'll verify once I get some time to do so.

src/services/response-service.ts
```diff
@@ -28,8 +31,7 @@ export default class ResponseService extends PercyClientService {
     }
 
     // Capture if the resourceUrl has a hostname in the allowedHostnames
-    const parsedResourceUrl = new URL(resourceUrl)
-    if (this.allowedHostnames.some((hostname) => parsedResourceUrl.hostname === hostname)) {
+    if (this.allowedHostnames.some((hostname) => domainMatch(hostname, resourceUrl))) {
```
wwilsman: The port number in the hostname causes domainMatch to fail. I just need to add back the `new URL` bit after lunch and this should work.

package.json
wwilsman: https://github.com/percy/percy-agent/pull/347#discussion_r325840590

wwilsman:
Unrelated, it looks like the CSSOM test flaked in one of the CI runs 🤔

Robdel12:
Which one?

Robdel12:
Oh weird, that looks like the live site failed to be requested or something 🤔

wwilsman:
The domain-match package had an issue where it was comparing hosts including the port number. Updated the approach (and PR description) to copy the small util over and update it to _not use the port number._

@Robdel12 could you re-review, including that new util file?

PR #349, opened 2019-09-19
feat: Capture redirected assets 

Robdel12:
## What is this? 

This PR introduces the ability to capture assets that are discovered, but have a redirect chain. Currently agent will skip these assets since it's not super easy to follow the chain all the way through and then capture that asset in Puppeteer.

## Approach

The redirected resources are captured by requesting the redirected URL with `axios` from Node. This will properly follow the redirect chain for us, then we save the response body of that request as the original request URL (so it proxy's correctly in our rendering env).

While doing this I refactored the methods for capturing resources to their respective paths now. Normal assets from the asset discovery service using puppeteer now process their resource and redirected resources are not processed via Node. 

## TODOs

- [x] Update the test SDK site with the redirect resource test in this PR (https://github.com/percy/sdk-test-website/pull/1)
- [x] `feat` or `fix`?

### DIFF
--- a/src/services/response-service.ts
+++ b/src/services/response-service.ts
@@ -80,7 +80,7 @@ export default class ResponseService extends PercyClientService {
     if (isRedirect) {
       // We don't want to follow too deep of a chain
       // `followRedirects` is the npm package axios uses to follow redirected requests
-      // we'll use their max redirect setting as a gaurd here
+      // we'll use their max redirect setting as a guard here
       if (request.redirectChain().length > followRedirects.maxRedirects) {
         logger.debug(`Skipping [redirect_too_deep: ${request.redirectChain().length}] [${width} px]: ${response.url()}`)
         return
--- a/src/services/response-service.ts
+++ b/src/services/response-service.ts
@@ -82,71 +82,99 @@ export default class ResponseService extends PercyClientService {
         return
       }
 
-      return this.handleRedirect(url, width, response, logger)
+      return this.handleRedirectResouce(url, response, width, logger)
     }
 
-    const localCopy = await this.makeLocalCopy(response, logger)
-    const responseBodySize = fs.statSync(localCopy).size
+    return this.handlePuppeteerResource(url, response, width, logger)
+  }
 
-    if (responseBodySize > Constants.MAX_FILE_SIZE_BYTES) {
-      // Skip large resources
+  /**
+   * Handle processing and saving a resource that has a redirect chain. This
+   * will download the resource from node, and save the content as the orignal
+   * requesting url. This works since axios follows the redirect chain
+   * automatically.
+   *
+   */
+  async handleRedirectResouce(redirectedURL: string, response: puppeteer.Response, width: number, logger: any) {
+    logger.debug(`Making local copy of redirected response: ${redirectedURL}`)
+
+    const { data } = await Axios(redirectedURL, { responseType: 'text' }) as any
+    const buffer = Buffer.from(data, 'utf8')
+    const sha = crypto.createHash('sha256').update(buffer).digest('hex')
+    const localCopy = path.join(os.tmpdir(), sha)
+    const didWriteFile = this.maybeWriteFile(localCopy, buffer)
+    const { fileIsTooLarge, responseBodySize } = this.checkFileSize(localCopy)
+
+    if (!didWriteFile) {
+      logger.debug(`Skipping file copy [already_copied]: ${response.url()}`)
+    }
+
+    if (fileIsTooLarge) {
       logger.debug(`Skipping [max_file_size_exceeded_${responseBodySize}] [${width} px]: ${response.url()}`)
       return
     }
 
+    // By not setting contentType, it serves it correctly in our proxy
+    const resource = this.resourceService.createResourceFromFile(redirectedURL, localCopy, '', logger)
+    this.responsesProcessed.set(redirectedURL, resource)
+
+    return resource
+  }
+
+  /**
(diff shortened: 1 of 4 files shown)


src/services/response-service.ts
```diff
@@ -12,7 +13,8 @@ import ResourceService from './resource-service'
 export default class ResponseService extends PercyClientService {
   resourceService: ResourceService
 
-  readonly ALLOWED_RESPONSE_STATUSES = [200, 201, 304]
+  readonly REDIRECT_STATUSES = [301, 302, 304, 307, 308]
+  readonly ALLOWED_RESPONSE_STATUSES = [200, 201, 301, 302, 304, 307, 308]
```
Robdel12: Nice way to combine these? Is it even worth it?

test/integration/agent-integration.test.ts
```diff
@@ -88,6 +88,13 @@ describe('Integration test', () => {
       expect(domSnapshot).contains('Buildkite')
     })
 
+    it('snapshots a site with redirected assets', async () => {
+      await page.goto('https://kind-goldstine-64b202.netlify.com/redirects/')
```
Robdel12: This needs to be updated when I update our SDK test site repo.

src/services/response-service.ts
wwilsman: Instance initializers are run in the context of the constructor (at the very end to be specific), so you have access to `this`.

```
readonly ALLOWED_RESPONSE_STATUSES = [200, 201, ...this.REDIRECT_STATUSES]
```

wwilsman: But I'm also not sure if these need to even be instance vars? If they don't change per instance and they aren't referenced by the instance outside of this file, couldn't these just be consts at the root level in this file?

src/services/response-service.ts
```diff
@@ -72,45 +75,106 @@ export default class ResponseService extends PercyClientService {
...
+    if (fileIsTooLarge) {
       logger.debug(`Skipping [max_file_size_exceeded_${responseBodySize}] [${width} px]: ${response.url()}`)
       return
     }
 
-    const contentType = response.headers()['content-type']
-    const resource = this.resourceService.createResourceFromFile(url, localCopy, contentType, logger)
-    this.responsesProcessed.set(url, resource)
+    // By not setting contentType, it serves it correctly in our proxy
+    const resource = this.resourceService.createResourceFromFile(redirectedURL, localCopy, '', logger)
```
wwilsman: I'm suspicious of not setting a content type 🤔 

Could we get the content type from the axios response?

src/services/response-service.ts
Robdel12: I agree, I’ll move it to a static const at the top

src/services/response-service.ts
Robdel12: Ah yes! I need to check this, but I might have to save the file as the actual file, then use a lib to get its mime type, then sha, then send. 

UNLESS you know of a better way where I can request with axios and the content type isn’t always `text/plain` (which then fails to serve in the proxy as css). Basically is there a better way to request the content of the file, and then turn it into a buffer?

wwilsman: Most libs use the file extension to determine the mime-type. So if you go that route, you might not have to write the actual filename.

As for axios, could you use `responseType: 'arraybuffer'`?

Following up with that, would `data` then already be a buffer?

src/services/response-service.ts
```diff
@@ -72,45 +75,107 @@ export default class ResponseService extends PercyClientService {
       return
     }
 
-    if (!this.shouldCaptureResource(rootUrl, response.url())) {
-      // Disallow remote redirects.
-      logger.debug(`Skipping [is_remote_redirect] [${width} px]: ${response.url()}`)
-      return
+    if (isRedirect) {
+      // We don't want to follow too deep of a chain
+      if (request.redirectChain().length > 4) {
```
Robdel12: I wonder if this even matters now that axios is doing all the work for us..?

Robdel12: Looks like the package that axios uses for this has a max set to 21 🤔 https://www.npmjs.com/package/follow-redirects#global-options

I'll update that to match. It would be real interesting to maybe open that to config _if we ever need to_

src/services/response-service.ts
```diff
@@ -72,45 +77,111 @@ export default class ResponseService extends PercyClientService {
...
+
+    if (fileIsTooLarge) {
+      logger.debug(`Skipping [max_file_size_exceeded_${responseBodySize}] [${width} px]: ${originalURL}`)
       return
     }
 
-    const contentType = response.headers()['content-type']
-    const resource = this.resourceService.createResourceFromFile(url, localCopy, contentType, logger)
-    this.responsesProcessed.set(url, resource)
+    // By not setting contentType, it serves it correctly in our proxy
```
wwilsman: ✂️ This comment doesn't apply anymore

src/services/response-service.ts
```diff
@@ -72,45 +77,111 @@ export default class ResponseService extends PercyClientService {
...
+      return true
+    } else {
+      return false
+    }
+  }
+
+  /**
+   * Ensures the saved file is not larger than what the Percy API accepts. It
+   * returns if the file is too large, as well as the files size.
+   *
```
wwilsman: Extra space under these descriptions 🤔

wwilsman:
Would this be a feature? Are we fixing redirects, or adding support for redirects?

🤷‍♂

PR #350, opened 2019-09-19
build: Correct `follow-redirects` dependency version

Robdel12:
## What is this?

This is so we're not depending on a newer version of `follow-redirects` than `axios` is. Now the same version of `follow-redirects` that axios uses is imported into our code (https://github.com/axios/axios/blob/v0.18.1/package.json#L74)

I don't think we need to release for this. It can go out with the next version 👍 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "cross-spawn": "^6.0.5",
     "deepmerge": "^4.0.0",
     "express": "^4.16.3",
-    "follow-redirects": "^1.9.0",
+    "follow-redirects": "1.5.10",
     "generic-pool": "^3.7.1",
     "globby": "^10.0.1",
     "image-size": "^0.8.2",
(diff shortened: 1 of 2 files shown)


wwilsman:
Looks like we're still resolving to 1.9.0.

Axios is pinned to 1.5.10 in your link

https://github.com/percy/percy-agent/blob/dabf2b26410d28bfaebe097dca498cce143606c4/yarn.lock#L4328

Robdel12:
@wwilsman forgot to reply to this! So, that's because karma uses `follow-redirects` too. That's only a dev dep, so if someone installs agent freshly it should resolve to the same as axios. 

Do you think I should hard pin it to the one that's hard pinned in axios?

wwilsman:
If it needs to match axios, let's pin it as well.

According to the lock file, our version is resolving to 1.9.0

https://github.com/percy/percy-agent/pull/350/files#diff-8ee2343978836a779dc9f8d6b794c3b2R4335-R4336

Robdel12:
@wwilsman fixed up 👍

PR #351, opened 2019-09-20
fix: Check for widths array length so empty arrays are never used

Robdel12:
## What is this?

If you pass an empty array from and SDK, the asset discovery service blows up. We should just fall back to the default widths


src/services/agent-service.ts
```diff
@@ -96,9 +96,10 @@ export class AgentService {
     const configuration = new ConfigurationService().configuration
     // trim the string of whitespace and concat per-snapshot CSS with the globally specified CSS
     const percySpecificCSS = configuration.snapshot['percy-css'].concat(request.body.percyCSS || '').trim()
+    const hasWidths = !!request.body.widths && request.body.widths.length
```
Robdel12: It's real annoying that `!![]` is true in JS

PR #352, opened 2019-09-24
fix: Add Percy specific CSS to root resource after asset discovery

Robdel12:
## What is this?

While debugging some percy CSS weirdness I thought it would be pretty nice to see in the logs when this is created and if its skipped, why?

### DIFF
--- a/src/services/agent-service.ts
+++ b/src/services/agent-service.ts
@@ -128,9 +128,9 @@ export class AgentService {
     )
 
     resources = resources.concat(
-      this.snapshotService.buildLogResource(snapshotLog),
       // @ts-ignore we won't write anything if css is not is passed
-      this.snapshotService.buildPercyCSSResource(percyCSSFileName, snapshotOptions.percyCSS),
+      this.snapshotService.buildPercyCSSResource(percyCSSFileName, snapshotOptions.percyCSS, snapshotLogger),
+      this.snapshotService.buildLogResource(snapshotLog),
     )
 
     const snapshotCreation = this.snapshotService.create(
--- a/src/services/agent-service.ts
+++ b/src/services/agent-service.ts
@@ -106,14 +106,7 @@ export class AgentService {
       minHeight: request.body.minHeight || configuration.snapshot['min-height'],
     }
 
-    let domSnapshot = request.body.domSnapshot
-    const percyCSSFileName = `percy-specific.${Date.now()}.css` as string
-
-    // Inject the link to the percy specific css if the option is passed
-    if (snapshotOptions.percyCSS) {
-      const cssLink = `<link data-percy-specific-css rel="stylesheet" href="/${percyCSSFileName}" />`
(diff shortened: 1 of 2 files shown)


src/services/agent-service.ts
```diff
@@ -130,7 +130,7 @@ export class AgentService {
     resources = resources.concat(
       this.snapshotService.buildLogResource(snapshotLog),
       // @ts-ignore we won't write anything if css is not is passed
-      this.snapshotService.buildPercyCSSResource(percyCSSFileName, snapshotOptions.percyCSS),
+      this.snapshotService.buildPercyCSSResource(percyCSSFileName, snapshotOptions.percyCSS, snapshotLogger),
```
wwilsman: The log has already been severed from the logger at this point. Move this line to before `buildLogResource` so they are still attached when you log inside the method.

Robdel12: Ah, right. Good catch

src/services/agent-service.ts
```diff
@@ -107,13 +107,6 @@ export class AgentService {
...
     let domSnapshot = request.body.domSnapshot
-    const percyCSSFileName = `percy-specific.${Date.now()}.css` as string
-
-    // Inject the link to the percy specific css if the option is passed
-    if (snapshotOptions.percyCSS) {
-      const cssLink = `<link data-percy-specific-css rel="stylesheet" href="/${percyCSSFileName}" />`
-      domSnapshot = domSnapshot.replace(/<\/body>/i, cssLink + '$&')
-    }
 
     if (domSnapshot.length > Constants.MAX_FILE_SIZE_BYTES) {
```
wwilsman: Is this limit a hard limit, or our own? If it's a hard limit, we might want to pull the `cssLink` var to above this so it can be used in the length calculation. `if ((domSnapshot.length + cssLink.length) > .....`

Robdel12: This is our own limit, just so logs weren't massive. I almost feel like we can _remove_ this now that we're collecting logs in a much different way. 

We did this because the DOM snapshot string would be MASSIVE and having people copy paste the logs became pretty hard.

wwilsman: If it isn't a hard limit, then it shouldn't matter that 80-100 bytes are added

src/services/snapshot-service.ts
```diff
@@ -23,27 +23,27 @@ export default class SnapshotService extends PercyClientService {
...
-    })
-
-    const discoveredResources = await this.assetDiscoveryService.discoverResources(
+    return this.assetDiscoveryService.discoverResources(
       rootResourceUrl,
       domSnapshot,
       options,
       logger,
     )
+  }
```
wwilsman: This method no longer seems useful. Could leave it as an alias, or just call `this.snapshotService.assetDiscoveryService.discoverResources` directly from the agent service.

Robdel12: Yeah for sure, going to clean that up in future PRs

PR #355, opened 2019-09-27
fix: Capture assets using the same hostname, regardless of port

Robdel12:
## What is this?

Currently, if someone is serving their app from `localhost:3001` and then serves their assets from `localhost:3002`, we will skip the assets coming from `localhost:3002`. This is because we're taking the port into account for asset discovery right now.

This change removes the port from consideration and captures all assets based on the `hostname` rather than `host`.

## TODO

- [ ] Verify this doesn't break proxying in weird cases.


Robdel12:
Going to close this since it's a stale PR. We can come back to it when ever we can verify it doesn't break anything on the infra side 👍

Issue #356, opened 2019-09-27
[Orb Enhancement] Add GitHub repo to orb description


KyleTryon:
Add `Repo: https://github.com/percy/percy-agent/` to orb description. 
https://github.com/percy/percy-agent/blob/master/.circleci/orb.yml

_Consider adding the `orbtoberfest` label on this issue to make this issue eligible for #Orbtoberfest https://circleci-public.github.io/hacktoberfest_

piyush-insider:
Hi, I would like to pick this up. Please add the `orbtoberfest` label to the issue.

djones:
@piyush-insider 👍 done

PR #363, opened 2019-09-30
docs: Added github repo to orb description

piyush-insider:
Added Github repo to Orb description
Resolves : #356

### DIFF
--- a/.circleci/orb.yml
+++ b/.circleci/orb.yml
@@ -1,7 +1,7 @@
 version: "2.1"
-description: Workflow for running Percy builds in parallel. Your token will be read from $PERCY_TOKEN by default, but this
-  can be overridden using the `percy_token` parameter.
-  Repo: https://github.com/percy/percy-agent/
+description: > 
+  Workflow for running Percy builds in parallel. Your token will be read from $PERCY_TOKEN by default, but this
+  can be overridden using the `percy_token` parameter. Source Code: https://github.com/percy/percy-agent/
 
 examples:
   finalize_all_percy_builds:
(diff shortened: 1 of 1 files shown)


piyush-insider:
@djones Can I please request your review on this?

djones:
Hi @piyush-insider, where is this documented with CircleCI? I don't see it in the configuration reference https://circleci.com/docs/2.0/configuration-reference/

piyush-insider:
Hi @djones It's listed as a best practice for metadata in the CircleCi documentation.
You can find it here.
https://circleci.com/docs/2.0/orbs-best-practices/#metadata

djones:
ok so it's part of the description field but this PR adds it as a YAML key on the next line, which ultimately won't get included in the description and actually causes a YAML validation error.

You can use `circleci orb validate` to check the configuration is valid. The validation comes back like this at the moment:

```bash
~/Code/percy-agent/.circleci(master ✗) circleci orb validate ./orb.yml
Error: Unable to parse YAML
mapping values are not allowed here
 in 'string', line 5, column 7:
      Repo: https://github.com/percy/percy ...
          ^
```

How about we change it to be something like this, using the `>` operator to get multiple lines working in the description.

```yaml
version: "2.1"
description: >
  Workflow for running Percy builds in parallel. Your token will be read from $PERCY_TOKEN by default, but this
  can be overridden using the `percy_token` parameter. Source code: https://github.com/percy/percy-agent

examples:
  finalize_all_percy_builds:
# ... rest of file
```

piyush-insider:
Hey, @djones thanks for pointing out the missing operator! I have updated the description accordingly.

Issue #357, opened 2019-09-27
[Orb Enhancement] Convert `percy_token` env var type to `env_var_name`


KyleTryon:
https://github.com/percy/percy-agent/blob/master/.circleci/orb.yml#L30

Utilize more secure parameter type `env_var_name` for environment variables.
https://circleci.com/docs/2.0/orbs-best-practices/#parameters

```
_Consider adding the `orbtoberfest` label on this issue to make this issue eligible for #Orbtoberfest https://circleci-public.github.io/hacktoberfest_
```

Shin-027:
Hi, I would like to pick this up.

KyleTryon:
#366 merged.

PR #366, opened 2019-10-01
build: change parameter type to env_var_name

Shin-027:
Fix issue #357 (Changed orb setting and description.)


Robdel12:
Thanks for the PR! 😃

Issue #359, opened 2019-09-27
[Orb Enhancement] Add missing job description.


KyleTryon:
The `finalize_all` job currently contains no description.

https://github.com/percy/percy-agent/blob/master/.circleci/orb.yml#L36

_Consider adding the `orbtoberfest` label on this issue to make this issue eligible for #Orbtoberfest https://circleci-public.github.io/hacktoberfest_

sccalabr:
Create pull request https://github.com/percy/percy-agent/pull/365. Let me know if I did anything wrong.

PR #368, opened 2019-10-01
Update orb.yml

Bhuvanesh01:
add missing job description


Robdel12:
Looks like #359 was merged which is the same change. Thanks for the PR though! 😃

PR #370, opened 2019-10-01
Change parameter type to env_var_name

Bhuvanesh01:


Robdel12:
Looks like this has the same change as #366

PR #383, opened 2019-10-01
feat: More config file options

wwilsman:
## Purpose

In part of making the configuration experience better by auto-loading `.env` files in percy-client, we can also expand `.percy.yml` support in agent to encompass more file types and use hierarchical priority!

## Approach

Uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) to search for various `.percy.*` config files and auto-loads said configuration. The library will start its search for the filenames in the current working directory and recurse upwards until a match is found.

There is also an option to load a specific config file by passing in a filename to the `loadSync` method. So targeted configs can become possible by allowing a flag to provide a filepath. I added this option under a `--config` flag and removed the Circle step to manually rename the `.ci.percy.yml` file.

I noticed the previous configurations service mutated the `DEFAULT_CONFIGURATION` object which causes a false positive test. I took this as an opportunity to improve how configuration is handled, and made it into a pure functional util. For flags and args parsing: instead of a multitude of `if` statements, a map is used and undefined values are filtered out.

`percy-client` was updated to include `.env` support. Also of note is cosmiconfig was already a subdependency, so there was no overhead adding it as our own dependency.


.circleci/config.yml
```diff
@@ -43,9 +43,6 @@ jobs:
       - run:
           name: Upload command test
           command: yarn test-upload-command
-      - run:
-          name: Setup integration test .percy.yml
-          command: mv .ci.percy.yml .percy.yml
```
Robdel12: 🧹🎉

PR #384, opened 2019-10-03
feat: Asset discovery headers

wwilsman:
## Purpose

The asset discovery service cannot capture assets protected by authentication. While headers may not help with cookie or session auth, they will at least enable users to use basic auth, proxy auth, or set some other header token to pass their server's auth requirements.

## Approach

Allow two options for specifying headers:

- Per-snapshot headers via a `requestHeaders` option – if only a single snapshot needs a header, this would be preferred over setting global request headers.

- Global headers via a `agent.asset-discovery.request-headers` option inside a config file.

Per-snapshot headers are merged with global headers. So headers defined in the config file will always be present, even when specifying additional headers per-snapshot.

The headers are set with Puppeteer's `setExtraHTTPHeaders()` method during asset discovery. For redirected assets, the original request's headers are forwarded to the Axios request.

### Tests

Live site tests cannot be added to the SDK test site due to Netifly restricting authentication to pro plans. The testing server used (http-server) recently added support for basic auth 6 months ago, but hasn't released a new version of the package for almost 2 years. That package was removed in favor of utilizing Express, which is already installed as a dependency, and `express-basic-auth` was added as a dev-dep for testing headers.


test/integration/agent-integration.test.ts
```diff
@@ -130,6 +134,30 @@ describe('Integration test', () => {
...
+        basicAuth({ users: { [username]: password }, challenge: true }),
+        express.static(testCaseDir),
+      )
+
+      await page.authenticate({ username, password })
+      await page.goto(`http://localhost:${PORT}/auth/protected-with-basic-auth.html`)
+
+      await snapshot(page, 'Protected assets', {
+        requestHeaders: {
+          Authorization: `Basic ${Buffer.from(`${username}:${password}`).toString('base64')}`,
```
Robdel12: What do you think about adding another snapshot that _doesn't_ pass the auth headers?

test/integration/testcases/protected-with-basic-auth.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Protected resources integration test case</title>
+    <meta charset="utf-8"/>
+  </head>
+  <body>
+    <p>All requests on this page are protected <img src="redirected.png" /></p>
+    <img src="10mb.png" />
```
Robdel12: Should we pick a smaller asset? haha

package.json
```diff
@@ -76,7 +75,7 @@
     "chai-http": "^4.0.0",
     "cheerio": "^1.0.0-rc.3",
     "eslint-config-oclif": "^3.0.0",
-    "http-server": "^0.11.1",
```
Robdel12: 👋 so long, you won't be missed (thinking back to all of the ecstatic issues on windows with this package)

test/integration/testcases/protected-with-basic-auth.html
wwilsman: Used the other emojione asset

test/integration/agent-integration.test.ts
wwilsman: Added. I actually tested this TDD style by making sure it failed before it passed 👍

PR #395, opened 2019-10-09
feat: :sparkles: Add client info for upload command

wwilsman:
Adds to the environment info section for builds when the upload command is used.

image

### DIFF
--- a/src/services/image-snapshot-service.ts
+++ b/src/services/image-snapshot-service.ts
@@ -92,6 +92,7 @@ export default class ImageSnapshotService extends PercyClientService {
       name,
       widths: [width],
       minimumHeight: height,
+      clientInfo: 'percy-upload',
     }).then(async (response: any) => {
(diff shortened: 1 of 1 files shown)


src/services/image-snapshot-service.ts
```diff
@@ -92,6 +93,7 @@ export default class ImageSnapshotService extends PercyClientService {
       name,
       widths: [width],
       minimumHeight: height,
+      clientInfo: `percy-upload/${pkg.version}`,
```
wwilsman: Is the package version necessary since we include the version of agent?

Robdel12: Probably not. 🤷‍♂

wwilsman: Removed the version since it's redundant

wwilsman: Removed the version since it's redundant

PR #396, opened 2019-10-09
fix: Always release puppeteer page in asset discovery

Robdel12:
## What is this?

There's a chance an error happens and then the page is never cleaned up. This
hangs the process open for ever.

### DIFF
--- a/src/services/asset-discovery-service.ts
+++ b/src/services/asset-discovery-service.ts
@@ -264,7 +264,7 @@ export class AssetDiscoveryService extends PercyClientService {
     await pool.release(page)
     profile('--> assetDiscoveryService.pool.release')
 
-    return maybeResources.filter((maybeResource) => maybeResource != null)
+    return maybeResources.filter(Boolean)
   }
 
   private async cleanPagePool() {
--- a/src/services/asset-discovery-service.ts
+++ b/src/services/asset-discovery-service.ts
@@ -237,6 +237,8 @@ export class AssetDiscoveryService extends PercyClientService {
       logger.debug(`Failed to load ${request.url()} : ${request.failure()!.errorText}}`)
     })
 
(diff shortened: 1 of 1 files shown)


src/services/asset-discovery-service.ts
```diff
@@ -247,23 +249,22 @@ export class AssetDiscoveryService extends PercyClientService {
...
-    return []
+    // always release the page from the pool
+    profile('--> assetDiscoveryService.pool.release', { url: rootResourceUrl })
+    page.removeAllListeners('request')
+    page.removeAllListeners('requestfinished')
+    page.removeAllListeners('requestfailed')
+    await pool.release(page)
+    profile('--> assetDiscoveryService.pool.release')
+
+    return maybeResources.filter((maybeResource) => maybeResource !== null)
```
wwilsman:
```suggestion
    return maybeResources.filter((maybeResource) => maybeResource != null)
```

src/services/asset-discovery-service.ts
```diff
@@ -247,23 +249,22 @@ export class AssetDiscoveryService extends PercyClientService {
...
-    return []
+    // always release the page from the pool
+    profile('--> assetDiscoveryService.pool.release', { url: rootResourceUrl })
+    page.removeAllListeners('request')
+    page.removeAllListeners('requestfinished')
+    page.removeAllListeners('requestfailed')
+    await pool.release(page)
+    profile('--> assetDiscoveryService.pool.release')
+
+    return maybeResources.filter((maybeResource) => maybeResource != null)
```
wwilsman: Total preference, but this could also be `.filter(Boolean)`

Robdel12: I like it. I'll update

wwilsman:
🎮

PR #399, opened 2019-10-15
Create requirements.txt

Golden-Bug:
creating requirements.txt file where people can add the requirements for their installation.


Robdel12:
Hi @Golden-Bug! I'm curious to know more about this change. What problem is this solving?

PR #400, opened 2019-10-16
feat: Allow passing in CHROMIUM_PATH environment variable

staffe:
I was trying to run Percy in our project’s current e2e test setup (running on Alpine, a minimal Docker image) but found I wasn’t able to change the Chromium path for the Puppeteer client in `percy-agent`, so I would like to suggest this solution to the problem.

See the issue described here: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker
> Getting headless Chrome up and running in Docker can be tricky. The bundled Chromium that Puppeteer installs is missing the necessary shared library dependencies.
> To fix, you'll need to install the missing dependencies and the latest Chromium package in your Docker file

An alternative solution would be to add the chromium path as an option in the Percy config or as a CLI option, but I would prefer it being an environment variable so it can be shared with the project’s puppeteer setup

### DIFF
--- a/src/services/asset-discovery-service.ts
+++ b/src/services/asset-discovery-service.ts
@@ -44,6 +44,7 @@ export class AssetDiscoveryService extends PercyClientService {
         '--no-sandbox',
         '--disable-web-security',
       ],
+      executablePath: process.env.CHROMIUM_PATH,
       ignoreHTTPSErrors: true,
       handleSIGINT : false,
     })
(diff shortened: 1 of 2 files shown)


wwilsman:
Hey @staffe!

I was looking at the Puppeteer docs again and I noticed that there might actually already be an env var that does this? Could you try setting [`PUPPETEER_EXECUTABLE_PATH`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#environment-variables) in your project?

staffe:
> Hey @staffe!
> 
> I was looking at the Puppeteer docs again and I noticed that there might actually already be an env var that does this? Could you try setting [`PUPPETEER_EXECUTABLE_PATH`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#environment-variables) in your project?

Thanks a lot, that fixed it!

PR #403, opened 2019-10-25
fix: start detached and stop cleanup

wwilsman:
## Purpose

An issue was brought up where the start and stop commands did not work as expected with the latest version of agent. In fixing that issue, I discovered that our test suite was not finishing and was exiting with a `0` status early.

After digging through CircleCI builds, I am unable to determine when these tests started to not be run in CI. It seem the very fist passing build has a warning about Percy exiting. Previous test runs cannot be viewed due to the junit mocha reporter only saving reports to a file.

## Approach

- Defaulted healthcheck helper argument to the default port and also made use of the healthcheck endpoint constant.

- Since flags no longer have defaults, we were passing the string `"undefined"` into the detached start process. This was fixed by conditionally setting those flags.

- The stop endpoint finalized the build, closed the server, and killed the running process. That last part caused builds to be re-finalized, therefore creating empty duplicate builds. This was fixed by cleaning up the PID file rather than outright killing the process. Since the server closes, the process will close anyway.

- Use the oclif `this.exit` method. Using `process.exit` caused the _mocha process_ to exit. Oclif exit stops script execution by throwing an error, so callbacks that referenced the stop method had to be adjusted.
  - Shared process handlers now pass a boolean to exit the process
  - The other call to stop in `exec` needed to be moved out of the callbacks and called after the subprocess exited. This subprocess was turned into a promise so that errors can be caught and we can always call close without creating an unhandled rejection.
  - The stdio test helpers were updated to ignore safe oclif exits

- All failing tests were fixed and updated. Even the skipped exec tests can be re-enabled. A couple of tests were removed due to them never working or being superseded by another change.

- Finally, junit was removed as the sole reporter so we can actually see the results of our tests in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ commands:
           name: Unit tests
           command: |
             mkdir -p reports
-            $NYC yarn test --reporter mocha-junit-reporter
+            $NYC yarn test
       - run:
           name: Client tests
           command: $NYC yarn test-client --singleRun
--- a/src/commands/exec.ts
+++ b/src/commands/exec.ts
@@ -58,11 +58,19 @@ export default class Exec extends PercyCommand {
     }
 
     // Even if Percy will not run, continue to run the subprocess
-    const spawnedProcess = spawn(command, argv, { stdio: 'inherit' })
-    spawnedProcess.on('exit', (code) => this.stop(code))
-    spawnedProcess.on('error', (error) => {
+    await new Promise((resolve, reject) => {
+      const spawnedProcess = spawn(command, argv, { stdio: 'inherit' })
+      spawnedProcess.on('exit', resolve)
+      spawnedProcess.on('error', reject)
+    }).catch((error) => {
+      // catch subprocess errors
       this.logger.error(error)
-      this.stop(1)
+      return 1
+    }).then((code: any) => {
+      // oclif exit raises an error to stop script execution, so the following
+      // stop method cannot be used inside of the subprocess event handlers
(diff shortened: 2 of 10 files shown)


.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ commands:
           name: Unit tests
           command: |
             mkdir -p reports
-            $NYC yarn test --reporter mocha-junit-reporter
+            $NYC yarn test
```
Robdel12: Do we lose the junit output entirely? It's nice to use Circles reporting from the junit reporter

Robdel12:
Maybe we should update the title to reflect better in the changelog? Something that includes fixing the `--detached` flag maybe?

.circleci/config.yml
wwilsman: Unfortunately with mocha, you can technically only have a single reporter. A junit file reporter for Circle failure reporting or being able to see successful CI output using the default reporter?

We could implement a third-party reporter like [mocha-multi-reporters](https://www.npmjs.com/package/mocha-multi-reporters), but I thought that could be another PR.

wwilsman:
This fixes more than just `--detached` though even those that's where this originated. It also fixes the agent server / `stop` command from finalizing a build twice, but I suppose other than that the other fixes are test-related.

.circleci/config.yml
Robdel12: Yeah that could be another PR. Wanna create a clubhouse / GH issue (whichever)?

PR #404, opened 2019-10-28
fix: parsing glob strings

### LINKED ISSUE #311: [Static site] Automatically strip spaces after commas
## What is this?

Currently if your comma separated list includes a space after the comma, the command will fail to work. This is probably true for all of the glob style options, but was specifically found for the 'ignore-files' option: https://github.com/percy/percy-agent/blob/master/src/services/static-snapshot-service.ts#L100-L103

We should make sure this isn't a gotcha for users for all of our options :)

wwilsman:
## Purpose

When passing a comma-separated list of globs, sometimes spaces might appear after commas causing the globs to start or end with a space.

I also noticed that valid globs with comma separated lists (commonly used for extensions) would be split incorrectly. i.e. `**/*.{html,htm}` and `**/*.{png,jpg,jpeg}`

Fixes #311 

## Approach

Created a common `parseGlobs` configuration utility that splits a comma-separated string, while preserving glob lists within `{...}`, and trims whitespace from glob patterns. It also performs the duty of filtering out empty strings.

The configuration unit tests were nested so this could be tested without the extra hooks used in the other config tests.

### DIFF
--- a/src/services/image-snapshot-service.ts
+++ b/src/services/image-snapshot-service.ts
@@ -7,6 +7,7 @@ import * as path from 'path'
 
 import { DEFAULT_CONFIGURATION } from '../configuration/configuration'
 import { ImageSnapshotsConfiguration } from '../configuration/image-snapshots-configuration'
+import { parseGlobs } from '../utils/configuration'
 import logger, { logError, profile } from '../utils/logger'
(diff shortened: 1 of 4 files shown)


Robdel12:
🏁 I manually finalized the Percy build

PR #419, opened 2019-11-08
feat: Introduce asset discovery response caching behind flag

Robdel12:
## What is this?

Sometimes the Percy Agent asset discovery service can overload a customers test server, which means sometimes snapshots are missing assets. Right now, we hit their applications test server for each asset at each width. 

Off by default, this PR introduces a way to enable response caching for asset discovery. This is so our asset discovery service only will hit their applications test server for a given asset once (if we get a 200 response). In theory, it should reduce the number of SDK related missing assets (most commonly seen in rails apps)

### DIFF
--- a/test/utils/response-cache.test.ts
+++ b/test/utils/response-cache.test.ts
@@ -37,7 +37,7 @@ describe('Response cache util', () => {
 
   it('calling the cache with the same URL does nothing', async () => {
     await cacheResponse(defaultResponse, logger)
-    await cacheResponse(defaultResponse, logger)
+    await cacheResponse({ ...defaultResponse, status() { return 201 } }, logger)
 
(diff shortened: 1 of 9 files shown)


test/utils/response-cache.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    expect(getResponseCache('http://example.com/foo.txt')).to.eql({
+      status: 201,
+      body: 'hello',
+      headers: 'fake headers',
+    })
+  })
+
+  it('calling the cache with the same URL does nothing', async () => {
+    await cacheResponse(defaultResponse, logger)
+    await cacheResponse(defaultResponse, logger)
```
wwilsman: We should use a different body with this second response to verify that only the very first response is cached

PR #420, opened 2019-11-11
build: set up initial acceptance tests

wwilsman:
## Purpose

It can sometimes be difficult to write high-level tests with the current test suites. To unit test a service you must mock out several other services. To test commands, the tests and commands themselves must be carefully written to avoid polluting the current process or exiting at all due to sharing the same process as the test suite (this recently caused us to ship quite a few bugs since tests weren't running at all).

This PR aims to set up an easier-to-write and more resilient acceptance suite, a suite that exists between our various unit tests and integration tests. These tests should run commands in their own processes via `sh` just as a user would run them, but they should not post back to the real API to create their own builds and snapshots. We can then write tests that assert calling `percy` in specific scenarios produces the correct and proper API requests.

These types of tests are very resilient to internal changes. For example, we could refactor the entire project away from OCLIF or swap object-oriented services for functional utils and use this test suite to verify if there are any breaking changes.

## Approach

A `run` helper uses `exec` with a deferred promise and subscriptions to the stdio streams are made to collect output. Similarly to the promisified version of exec, the process is added to the promise via a `child` property, however the promisified version is not used so we can strip colors from the output to test easier, and still resolve the promise even when the subprocess exits with a non-zero. 

When using the `run` helper, new matchers are added to `expect` to assert against ordered entries within the output: `.toHaveEntry(match)` and `.toHaveEntries(match[])` where `match` is a string, substring, or regular expression matching an entry in the output array. `.toHaveEntries` will also ensure that each entry appears in the same order as the provided array of matches.

To capture API requests, I first explored the generic `HTTPS_PROXY` environment variable which seemed promising but became difficult to respond with mocks. Instead, I utilized the `PERCY_API` environment variable which tells the underlying client to use a different API URL. In these tests, this API URL routes to an express server where we can assert against requests and respond with mocks. A `proxy` helper will set up the server and some default mocks. It also contains a `mock` method to override the default mocks for specific scenarios. The proxy server will capture every request made to it and assertions can be made against those requests after the `percy` command has finished.

[This test is a good example of using both the `run` and `proxy` helpers to assert that the build is created before the command runs, and then finalizes when the command finishes](https://github.com/percy/percy-agent/pull/420/files#diff-6eb0f9145b7b7ede4e7fbf0d7954e346R19-R38)

A new dummy app was added to the acceptance tests. Two helpers come with this dummy app: `setupDummyApp` and `launch`. The `setupDummyApp` helper will hook into mocha beforeEach and afterEach to start and stop the dummy app server. This express app may also be interacted with in tests to set up custom routes or authentication. The `launch` helper is a small reimplementation of PercyScript for this test suite (`launch` was chosen because `run` is already used by another helper). The difference is – any errors are **not** swallowed and/or logged and are instead allowed to bubble into the test so that errors can be surfaced and debugged.


tests/helpers/snapshot.js
```diff
@@ -0,0 +1,58 @@
...
+      '--disable-gpu',
+      '--no-sandbox',
+      '--single-process',
+      '--disable-dev-profile'
+    ]
+  })
+
+  try {
+    let page = await browser.newPage()
+    await page.setBypassCSP(true)
```
Robdel12: Are we sure we want to set this here? Are there cases where we'll want CSP bypassing off?

package.json
```diff
@@ -75,7 +77,10 @@
     "chai-http": "^4.0.0",
     "cheerio": "^1.0.0-rc.3",
     "eslint-config-oclif": "^3.0.0",
+    "expect": "^24.9.0",
```
Robdel12: Future PR to remove chai (and code mod to `expect`)?

tests/helpers/snapshot.js
wwilsman: Without it, I can't even inject the client script into this local express server. I figured if it's used in PercyScript, probably should just use it here in this mini-clone.

package.json
wwilsman: Yes! Custom matchers in `expect` are so much easier to write, and you don't have to deal with weird Chai idiosyncrasies like `.deep.equal` or `.deep.contains` only performing partial matching at the top-level, or even just the confusing nature of "_which chai assertion should I use and which order should the `.confusing.extended.syntax.go`?_"

tests/helpers/snapshot.js
Robdel12: Ah, gotcha. Makes sense

test/acceptance/helpers/snapshot.js
```diff
@@ -0,0 +1,81 @@
...
+ * @returns {string} The captured DOM string
+ */
+async function snapshot(page, name, options = {}) {
+  await page.addScriptTag({
+    url: 'http://localhost:5338/percy-agent.js'
+  })
+
+  let dom = await page.evaluate(options => {
+    return new PercyAgent({ handleAgentCommunication: false })
+      .domSnapshot(document, options)
```
Robdel12: Very nice. I think we should explore exporting the DOM class and not even using `percy-agent.js` anymore. All of the SDKs do this exact thing, which only passes these args down to `new DOM`

test/acceptance/helpers/snapshot.js
```diff
@@ -0,0 +1,81 @@
...
+  }, options)
+
+  let data = JSON.stringify({
+    name,
+    url: page.url(),
+    domSnapshot: dom,
+    clientInfo: 'acceptance-micro-sdk',
```
Robdel12: Wonder if we should also pass any node env info?

test/unit/.mocha.yml
```diff
@@ -0,0 +1,8 @@
+require:
+  - "test/unit/helpers/init.ts"
+  - "ts-node/register"
+  - "source-map-support/register"
+watch-extensions: "ts"
+recursive: true
+reporter: "spec"
+timeout: 5000
```
Robdel12: 👍 Nice

test/acceptance/helpers/snapshot.js
```diff
@@ -0,0 +1,81 @@
...
+ * assertions within tests.
+ *
+ * @param {object} page - Puppeteer page objectg
+ * @param {string} name - Snapshot name
+ * @param {object} options - Snapshot options
+ * @returns {string} The captured DOM string
+ */
+async function snapshot(page, name, options = {}) {
+  await page.addScriptTag({
+    url: 'http://localhost:5338/percy-agent.js'
```
Robdel12: I quite like this. We could easily remove the SDK util in favor of this for other SDKs

test/acceptance/helpers/run.js
```diff
@@ -0,0 +1,99 @@
...
+      message: () => [
+        `${this.utils.matcherHint('toHaveEntries', { isNot: this.isNot })}\n`,
+        `Expected output to ${this.isNot ? 'not ' : ''}contain entries (in order):`,
+        `  ${this.utils.printExpected(expected)}`,
+        'Received:',
+        `  ${this.utils.printReceived(actual)}`
+      ].join('\n')
+    }
+  }
+})
```
Robdel12: These matchers are going to be _real_ nice for writing tests. Not enough credit is given to good tooling 👍

test/acceptance/helpers/dummy.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ * Hooks into Mocha's `beforeEach` and `afterEach` to create and close the dummy
+ * server before and after each test. A local context is mutated and cleaned up
+ * to handle memory leaks and returning helpers for use in tests.
+ *
+ * @returns {object} An object containing app and server references
+ */
+export default function setupDummyApp(options) {
```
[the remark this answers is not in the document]
wwilsman: Not sure what you mean by 'embed' in this instance. Like embed the netifly site itself? Or just move those pages over into this dummy app? 

My concern with the former would be that you need to make and merge changes to an entirely different repo just to write a test. With this dummy app, I'd like to make sure all tests stay within this repo. So we don't hit _any_ live site. Not even the test site (cause that's still a live site that would cause our tests to fail if any internet issues arise).

test/acceptance/helpers/snapshot.js
wwilsman: We don't have per-snapshot client info do we? Otherwise our merged builds would show both node versions in the client info section.

Although, these tests never send things off to the real API, so this client info is probably not even necessary. I'll keep it just in case it comes in handy in the future.

test/acceptance/helpers/snapshot.js
wwilsman: The possible caveat is CSP issues. I had to bypass it to inject this script into the dummy express app.

test/acceptance/helpers/snapshot.js
Robdel12: > Although, these tests never send things off to the real API, so this client info is probably not even necessary. I'll keep it just in case it comes in handy in the future.

Hahaha, very very true. Good point, forgot that. In the real API, we store per-snapshot user agents I believe. Just not surfaced in the UI

PR #421, opened 2019-11-15
fix: Add better logging for command not found errors

### LINKED ISSUE #297: Gracefully exit with a nice error if we can't find the passed command
It's not uncommon for folks to reach out to support with an error that looks like: `Error: spawn cypress ENOENT`

This is usually because the passed command to `percy exec --` can't be found. We should capture this error and show a nice error message that says the command can't be found. Maybe link to a doc nudging the user to create a script in their `package.json` so the binary can be found.

wwilsman:
## Purpose

We started noticing `[percy] undefined` in the logs whenever the command being executed was not found in the $PATH. This was due to the error object being passed through from the child's `error` handler being interpreted incorrectly by Winston 3.x. Winston attempts to find a `msg` property on the passed object, which is where `undefined` was coming from. Winston does have logic for handling Error instances, but for some reason the error provided by cross-spawn did not get handled like a normal Error.

Fixes #297 

## Approach

Call `.toString()` on the error to log the error message rather than the error object.

The error this produced was `spawn foobar ENOENT` which might not be obvious to some that it means the command was not found. So I utilized a transitive dependency `which` to perform a cross-platform check for the provided command so we can log a much friendlier error.

In addition, I added a non-zero exit code to the missing command logic and tested that as well.

### DIFF
--- a/src/commands/exec.ts
+++ b/src/commands/exec.ts
@@ -1,5 +1,6 @@
 import { flags } from '@oclif/command'
 import * as spawn from 'cross-spawn'
+import * as which from 'which'
 import { DEFAULT_CONFIGURATION } from '../configuration/configuration'
 import config from '../utils/configuration'
 import PercyCommand from './percy-command'
@@ -56,7 +57,12 @@ export default class Exec extends PercyCommand {
       this.logger.info('You must supply a command to run after --')
       this.logger.info('Example:')
       this.logger.info('$ percy exec -- echo "run your test suite"')
-      return
+      return this.exit(1)
+    }
+
+    if (which.sync(command, { nothrow: true }) == null) {
+      this.logger.error(`Error: command not found "${command}"`)
+      return this.exit(127)
     }
 
     if (this.percyWillRun()) {
@@ -70,7 +76,7 @@ export default class Exec extends PercyCommand {
       spawnedProcess.on('error', reject)
     }).catch((error) => {
       // catch subprocess errors
-      this.logger.error(error)
+      this.logger.error(error.toString())
       return 1
     }).then((code: any) => {
       // oclif exit raises an error to stop script execution, so the following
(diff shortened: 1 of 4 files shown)


src/commands/exec.ts
```diff
@@ -1,5 +1,6 @@
 import { flags } from '@oclif/command'
 import * as spawn from 'cross-spawn'
+import * as which from 'which'
```
Robdel12: Do you know if this is cross platform safe? I couldn't see anything that says yay or nay on the packages npm page

test/acceptance/exec.test.js
```diff
@@ -18,6 +18,26 @@ describe('percy exec', () => {
...
+      '[percy] Example:',
+      '[percy] $ percy exec -- echo "run your test suite"'
+    ])
+  })
+
+  it('errors when the command cannot be found', async () => {
+    let [stdout, stderr, code] = await run('percy exec -- foobar')
+
+    expect(code).toEqual(127)
+    expect(stderr).toHaveEntry('[percy] Error: command not found "foobar"')
```
Robdel12: I LOVE these tests hahaha

src/commands/exec.ts
Robdel12: https://github.com/npm/node-which/blob/8ef60698398872699b5446a06fc579d281a94b91/which.js#L6 looks like it does.

Totally want to verify this on a windows machine before though 😬

wwilsman: It's a dependency of cross-spawn, so it is verified to work on a windows machine 🙂

wwilsman: I initially used `spawn.sync('which', [command])` but added the dep for winders

wwilsman: Also from the PR description 😜 

> So I utilized a transitive dependency which to perform a cross-platform check

PR #423, opened 2019-11-21
fix: Wait for asset discovery finish when awaiting on snapshot resources in finalize

Robdel12:
## What is this? 

The title is a mouthful, that's for sure. If the child process spawned by agent in `exec` exits while snapshots are processing, it's possible for snapshots to be added to a _finalizing_ build. This will break the rendering pipeline for that build. 

## Approach

First up was to write an acceptance test that reproduces this behavior. The easiest way would be to have a PercyScript file that loops through snapshots and throws an exception while taking snapshots. Then assert to make sure no snapshots were added to to the build _after_ we've finalized the build. 

Quick solution: As soon as `handleSnaphshot` is called, add a promise to the `snapshotCreationPromises` array. That way any deferred work can still be awaited on (when we're stopping agent).

Better (future) solution: After finalize is called for a build, no more snapshot or resource requests should be uploaded. This might require more of a sweeping change to how services are interconnected yet decoupled from command classes.

## TODOs

- [x] Assert on `waiting for x snapshots` stdout 
- [x] Investigate integration test issues

### DIFF
--- a/test/acceptance/asset-discovery.test.js
+++ b/test/acceptance/asset-discovery.test.js
@@ -12,7 +12,7 @@ describe('Asset discovery', () => {
   it('does not create snapshots while stopping a build', async () => {
     let [stdout] = await run('percy exec -- node ./test/acceptance/dummy/snapshot-error.js');
 
-    // TODO: Assert the "waitig for x" output & for no snapshot logs to occur *after*
+    // TODO: Assert the "waiting for x" output & for no snapshot logs to occur *after*
     expect(stdout).toHaveEntries([
       '[percy] percy has started.',
       "[percy] snapshot taken: 'Home Page - 0'",
--- a/test/acceptance/asset-discovery.test.js
+++ b/test/acceptance/asset-discovery.test.js
@@ -16,6 +16,7 @@ describe('Asset discovery', () => {
   it('does not create snapshots while stopping a build', async () => {
     let [stdout] = await run('percy exec -- node ./test/acceptance/dummy/snapshot-error.js');
 
+    // TODO: Assert the "waitig for x" output & for no snapshot logs to occur *after*
     expect(stdout).toHaveEntries([
       '[percy] percy has started.',
       "[percy] snapshot taken: 'Home Page - 0'",
--- a/test/acceptance/asset-discovery.test.js
+++ b/test/acceptance/asset-discovery.test.js
@@ -1,0 +1,26 @@
+import expect from 'expect'
+import { execSync } from 'child_process'
+import { when } from 'interactor.js'
+
+import {
+  launch,
+  run,
+  setupApiProxy,
+  setupDummyApp
+} from './helpers'
+
+describe('Asset discovery', () => {
+  let proxy = setupApiProxy()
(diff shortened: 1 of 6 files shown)


test/acceptance/asset-discovery.test.js
```diff
@@ -0,0 +1,27 @@
...
+  run,
+  setupApiProxy,
+  setupDummyApp
+} from './helpers'
+
+describe('Asset discovery', () => {
+  let proxy = setupApiProxy()
+  let dummy = setupDummyApp()
+
+  it('does not create snapshots while stopping a build', async () => {
```
wwilsman: More accurately: `waits for snapshot finalization before build finalization`

test/acceptance/asset-discovery.test.js
```diff
@@ -0,0 +1,27 @@
...
+    let [stdout] = await run('percy exec -- node ./test/acceptance/dummy/snapshot-error.js');
+
+    let finalizeReqDate = proxy.requests['/builds/123/finalize'][0].requestDate;
+    let lastSnapshotReqDate = proxy.requests['/builds/123/snapshots'].pop().requestDate;
```
wwilsman: I dislike this mutation, but the proxy is reset before each test, so my disdain isn't valid.

test/acceptance/asset-discovery.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+    expect(finalizeReqDate > lastSnapshotReqDate).toBe(true)
```
wwilsman:
```suggestion
    expect(finalizeReqDate).toBeGreaterThan(lastSnapshotReqDate)
```

test/acceptance/helpers/proxy.js
```diff
@@ -50,7 +50,7 @@ export function createApiProxy() {
   app.all('*', (req, res, next) => {
     let path = req.path.replace(/\/$/, '')
     requests[path] = requests[path] || []
-    requests[path].push(req)
+    requests[path].push({ requestDate: new Date(), ...req })
```
wwilsman: These `Request` objects already live in a `requests` map, it's probably redundant to call `request.requestDate`. Probably also doesn't need to be a full Date object. What do you think of `{ timestamp: Date.now(), ...req }`?

test/acceptance/asset-discovery.test.js
wwilsman: Snapshot creation*? I guess technically we want to make sure both creation and finalization happen before the build finalizes

Robdel12: I like it 👍 We don't _have_ to finalize the snapshot right now (the API doesn't require it). But it is nice to do that for perf

test/acceptance/asset-discovery.test.js
Robdel12: Its valid, I just wanted a clean way to get the last item. I'll not mutate

test/acceptance/helpers/proxy.js
Robdel12: Love it :shipit:

PR #433, opened 2019-12-02
build: start command acceptance

wwilsman:
## Purpose

Sets up initial acceptance tests for the `start` command

## Approach

Updated the `run` helper to accept an additional argument to enable piping output to the current process for debugging. A final arg of `true` to the `run` helper will set up the pipe for the parent process. The sdio of the child process is still captured during debugging.

Added two basic acceptance tests for `start`:

- A warning is logged when there is no PERCY_TOKEN
- The server stops and finalizes the build when the process is killed

I initially added a third test, to test that `--detached` works as expected. However, I discovered that it will save log info to a hardcoded path in the cwd, which isn't ideal for testing. Even if this was made configurable, using `--detached` has other consequences that prevent it from connecting with the local test proxy (haven't debugged fully). Finally, while attempting to test this, I fixed an issue where the PID file was never properly cleaned up if the process was killed.

Given all of the issues with `--detached`, it might be better to just recommend using the built-in `&` to run processes in the background. Logs can be easily capture by adding `> some-file.log` to the CLI. The only change needed to support `&` would be to remove the PID file completely and have the `stop` command do a health check rather than check for a local PID file.

#### Unrelated:

Also added a test for the exec exit code when PERCY_TOKEN is missing

### DIFF
--- a/test/acceptance/start.test.js
+++ b/test/acceptance/start.test.js
@@ -18,15 +18,14 @@ describe('percy start', () => {
   })
 
   it('starts the agent server and stops it when the process is killed', async () => {
-    let start = run('percy start')
-
+    let start = run('percy start', true)
     // kill the process as soon as it's ready
-    await agentIsReady()
-    // puppeteer will complain if we kill the process while it is launching chrome; wait 100ms
-    setTimeout(() => start.child.kill(), 100)
+    await when(() => expect(start.stdio[1]).toHaveEntry('percy is ready.'), 3000)
+    require('child_process').execSync(`kill ${start.child.pid}`)
 
-    let [stdout, stderr] = await start
+    let [stdout, stderr, code] = await start
 
+    console.log('finished!', code)
     expect(stderr).toHaveLength(0)
     expect(stdout).toHaveEntries([
       '[percy] created build #4: <<build-url>>',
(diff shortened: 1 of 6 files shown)


test/acceptance/start.test.js
```diff
@@ -0,0 +1,37 @@
...
+    let [stdout, stderr, code] = await run('percy start', { PERCY_TOKEN: '' })
+    expect(stderr).toHaveEntry('Warning: Skipping visual tests. PERCY_TOKEN was not provided.')
+    expect(stdout).toHaveLength(0)
+    expect(code).toBe(0)
+  })
+
+  it('starts the agent server and stops it when the process is killed', async () => {
+    let start = run('percy start')
+    // kill the process 500ms after it creates a build
+    await when(() => !!proxy.requests['/builds'].length, 3000)
```
Robdel12: Looks like this is failing in CI 👀

wwilsman: Not so much that this was failing, but that the following kill was failing because agent wasn't really running yet. Since kill failed, the `await start` made the test timeout.

`when` was too fast and even though the build was created, the server hadn't been started. Fixed by using a more determinstic helper, but still might flake because of puppeteer launching Chrome (even though this test doesn't send snapshots, we launch chrome on server init).

wwilsman: Looks like it's still failing to kill agent 🤔

Robdel12:
Closing due to staleness 👍

PR #435, opened 2019-12-11
fix: don't log success on error

wwilsman:
## Purpose

Even if the finalize request fails, we currently print a `finalized build` message (which is not true)

## Approach

@Robdel12 did the fix, I added a test and sent the error to stderr rather than stdout.

Previously, we were `catch`ing a promise and _then_ logging "successfully." Since the promise thinks the error is "handled" it allows execution to continue. Moving `catch` to be after `then` rather than before will correctly log on a successful response only. Also, rather than mixing `async`/`await` and `then/catch`, this change strictly sticks to `async`/`await` (and uses a `try`/`catch` block instead of promise methods).

### DIFF
--- a/src/services/build-service.ts
+++ b/src/services/build-service.ts
@@ -35,7 +35,7 @@ export default class BuildService extends PercyClientService {
       await this.percyClient.finalizeBuild(this.buildId)
       this.logEvent('finalized')
     } catch (err) {
-      console.log(err)
+      logError(err)
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/services/build-service.ts
```diff
@@ -31,8 +31,12 @@ export default class BuildService extends PercyClientService {
   async finalize() {
     if (!this.buildId) { return }
 
-    await this.percyClient.finalizeBuild(this.buildId).catch(logError)
-    this.logEvent('finalized')
+    try {
+      await this.percyClient.finalizeBuild(this.buildId)
+      this.logEvent('finalized')
+    } catch (err) {
+      logError(err)
```
timhaines: Can we add some context here..  Perhaps a log line above this logError that says `Error during build finalize:` or something similar.

PR #437, opened 2019-12-13
fix: skip caching empty resources

blake-newman:
skip caching empty resources (browser resource hints)


Robdel12:
Hey @blake-newman, thanks for the PR! Could you add an acceptance test for this too? https://github.com/percy/percy-agent/blob/master/test/acceptance/exec.test.js

If not, we can get to it next week 👍

wwilsman:
Merged with #458

PR #439, opened 2019-12-18
fix: remove TypeScript types from dependencies

alexlafroscia:
I noticed when installing `@percy/agent` that my `yarn.lock` suddenly had some TypeScript definitions in it, despite me not using TypeScript.

I believe that the `@types/which` package should actually be a `devDependency`, rather than a `dependency`. Otherwise, all of your users have to install these type definitions along with your package!


taion:
TypeScript explicitly encourage listing type definitions for runtime dependencies under `dependencies` rather than `devDependencies`, BTW: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies

The reason being that otherwise consumers of your package who do use TypeScript don't get the type definitions installed, and won't be able to use any types from your dependencies that you expose.

The dependency is essentially "free" for people who don't use TypeScript, as the code ends up just sitting in `node_modules` not doing anything. It doesn't show up at runtime at all.

PR #453, opened 2020-01-13
fix: Ensure static image sizes are explicit

wwilsman:
## Purpose

If a browser has a minimum width it is willing/able to open at, statically uploaded images are set to be 100% of that width. This can sometimes cause the image to be cropped.

## Approach

Set a max-width so images don't exceed past the edges of the browser, but be explicit about setting the appropriate width and height attributes. This should still be rendered at 100% width of the browser, but in cases where the browser cannot be as narrow as the image, this should prevent cropping.


wwilsman:
Looks like some things like `caniuse-lite` might pollute the stderr which causes some acceptance tests to fail. In the future, we should look into silencing logs from other libraries. For now, I upgraded the package.

Robdel12:
Looks like it's still failing. Go away caniuse! lol

PR #458, opened 2020-01-21
fix: skip caching empty resources

wwilsman:
## Purpose

Builds onto #437 by adding tests

## Approach

Added a unit test for this since our acceptance tests might be a little flakey due to slow startup times.

This suite was adjusted to account for now requiring a couple other response methods.

Also in this test suite, I replaced instances of `() { return 'foo' }` with `() => 'foo'` except where we explicitly need to reference `this`.

---

Skipped a live site test due to the response being very slow. We have a ticket to remove (and replace if necessary) these live tests, but that should be in another PR. Skipping the failing one for now rather than increasing the timeout to 20 seconds.

### DIFF
--- a/test/integration/agent-integration.test.ts
+++ b/test/integration/agent-integration.test.ts
@@ -78,12 +78,6 @@ describe('Integration test', () => {
       expect(domSnapshot).contains('badssl.com')
     })
 
-    it('snapshots a complex website with responsive images', async () => {
-      await page.goto('https://polaris.shopify.com/')
-      const domSnapshot = await snapshot(page, 'Polaris snapshot', {widths: [300, 1200]})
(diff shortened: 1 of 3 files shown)


test/integration/agent-integration.test.ts
```diff
@@ -78,7 +78,7 @@ describe('Integration test', () => {
       expect(domSnapshot).contains('badssl.com')
     })
 
-    it('snapshots a complex website with responsive images', async () => {
+    it.skip('snapshots a complex website with responsive images', async () => {
```
Robdel12: I think we can just delete this test tbh. We have a responsive image test further down that isn't a live site.

wwilsman: 💯

Issue #459, opened 2020-01-31
Uploading large images with `percy upload`


danielschleindlsperger:
Hey!

I just stumbled upon the problem where I want to upload a large image with `percy upload` and I'm getting the error
```sh
[percy] StatusCodeError 400 - {"errors":[{"status":"bad_request","detail":"The minimum-height attribute must be between 10 and 2000 inclusive. Received 6465."}]} | Fri Jan 31 2020 18:26:48 GMT+0100 (Central European Standard Time)
```

I think percy-agent sets the minimum-height based on the height of the image, which then fails backend validation.

I can't upload an HTML snapshot, since some of the assets are on my company's private staging system behind a firewall. So I figured I would just manually take the screenshots and upload them to percy for the diffing and approval workflow goodies.

Maybe I'm doing something wrong or there is another workaround I'm not thinking of?

Thanks in advance and have a nice weekend!

Robdel12:
Hey Daniel,

Thanks for the report! I'm going to side step the `upload` stuff for a second and see if we can address this first (and then maybe we don't have to worry about `upload`). 

> I can't upload an HTML snapshot, since some of the assets are on my company's private staging system behind a firewall. So I figured I would just manually take the screenshots and upload them to Percy for the diffing and approval workflow goodies.

We support [capturing many asset hosts](https://docs.percy.io/docs/capturing-multiple-asset-hosts) in our SDKs, I'd be curious if that would work for you here. It sounds like the exact use case we built it for. 😃 What that will do is save those assets from the hosts you specify so they're locally available in our rendering env.

danielschleindlsperger:
Thanks for the speedy response! 

Capturing asset hosts would actually work perfectly if our site were fully responsive 😅 
We're already using Percy with Storybook to responsively test components (although we had to use another workaround here since I don't belive `percy-storybook` uses the sdk configuration).
Right now I'm trying to snapshot full pages for mobile, desktop and AMP (each with specific viewport width, user-agent and url)which doesn't quite fit with the responsive viewport `widths` model of Percy.

Anyway it seems Wil has already fixed it in the CLI 😁 🙏

wwilsman:
PR was just merged. 😃 

Should be released now as v0.20.13

danielschleindlsperger:
Perfect, thank you so much! 🙏

PR #460, opened 2020-01-31
fix: clamp image dimensions to min and max for snapshot width and minHeight options

wwilsman:
## Purpose

#459 brought to light that sometimes image dimensions might be outside of the bounds of our min and max for snapshots.

## Approach

Clamp both width and height to between 10px - 2000px, but for snapshot options `widths` and `minHeight` only as the image in the DOM should contain the correct width and height.

When we render on our side, the full image should still be captured.

### TODO

- [ ] Should I add a matrix of images to test this? 
  - [ ] Image width < 10px
  - [ ] Image width > 2000px
  - [ ] Image height < 10px
  - [ ] Image height > 2000px
  - [ ] Image width < 10px, height < 10px
  - [ ] Image width < 10px, height > 2000px
  - [ ] Image width > 2000px, height < 10px
  - [ ] Image width > 2000px, height > 2000px

### DIFF
--- a/src/services/image-snapshot-service.ts
+++ b/src/services/image-snapshot-service.ts
@@ -93,8 +93,9 @@ export default class ImageSnapshotService extends PercyClientService {
   ): Promise<any> {
     return this.percyClient.createSnapshot(this.buildId, resources, {
       name,
-      widths: [width],
-      minimumHeight: height,
+      // clamp between 10px - 2000px
+      widths: [Math.max(10, Math.min(width, 2000))],
+      minimumHeight: Math.max(10, Math.min(height, 2000)),
       clientInfo: 'percy-upload',
     }).then(async (response: any) => {
       await this.percyClient.uploadMissingResources(this.buildId, response, resources)


wwilsman:
Hmmm... why are there new snapshots and a missing one? They should all be there :thinking:

Robdel12:
Not sure why the logo is missing, but the new snapshots are because the width is now 812px

src/services/image-snapshot-service.ts
```diff
@@ -93,8 +93,9 @@ export default class ImageSnapshotService extends PercyClientService {
   ): Promise<any> {
     return this.percyClient.createSnapshot(this.buildId, resources, {
       name,
-      widths: [width],
-      minimumHeight: height,
+      // clamp between 10px - 2000px
+      widths: [Math.max(10, Math.min(height, 2000))],
```
wwilsman: Lol
```suggestion
      widths: [Math.max(10, Math.min(width, 2000))],
```

PR #468, opened 2020-02-21
feat: iframe serialization

wwilsman:
## Purpose

Currently, we only capture iframes during the asset discovery process. In which we only capture the response of the `src` attribute. If an iframe has been interacted with, navigated within, or built with JavaScript, we cannot properly capture that iframe.

This PR aims to solve those limitations

## Approach

During DOM serialization, we can retrieve the DOM of an iframe using `iframe.contentDocument`. We can then recursively serialize that DOM, and once serialized set the content as a `srcdoc` attribute. The `src` attribute is then removed since it is no longer needed and might cause other issues.

While dealing with iframes, we can also remove iframes from the `<head>`. These iframes typically have no visual presentation and when they do, they often break pages since rerendering causes the browser to "fix" the `<head>` and pushes everything from the iframe down into an implicit `<body>` (and ignore the explicit `<body>`).

The `contentDocument` property returns `null` for iframes with security restrictions (CORS), so we have no way to serialize those iframes. We also do not want to serialize iframes that have not finished loading (even with CORS, you can access the DOM before it loads and incorrectly serialize an empty DOM). Firefox does not trigger load-related properties for iframes built with JS, so those iframes are always serialized (as long as JS is not enabled).

There are simple tests included here, but as a complex example I manually tested this on a live site where the iframe has internal navigation on form submission. [Results here](https://percy.io/percy/wil-test-bed/builds/4254676).

### DIFF
--- a/src/percy-agent-client/dom.ts
+++ b/src/percy-agent-client/dom.ts
@@ -93,6 +93,7 @@ class DOM {
    */
   private stabilizeDOM(clonedDOM: HTMLDocument): HTMLElement {
     this.serializeInputElements(clonedDOM)
+    this.serializeFrameElements(clonedDOM)
 
     // We only want to serialize the CSSOM if JS isn't enabled.
     if (!this.options.enableJavaScript) {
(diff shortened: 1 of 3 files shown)


test/unit/percy-agent-client/dom.test.ts
```diff
@@ -328,5 +328,65 @@ describe('DOM -', () => {
...
+          <iframe id="frame-external" src="https://example.com"></iframe>
+          <iframe id="frame-input" srcdoc="<input/>"></iframe>
+          <iframe id="frame-js" src="javascript:void(
+            this.document.body.innerHTML = '<p>made with js src</p>'
+          )"></iframe>
+          <iframe id="frame-js-no-src"></iframe>
+        `)
+
+        const $frameInput = document.getElementById('frame-input') as HTMLIFrameElement
+        await when(() => !!$frameInput.contentWindow!.performance.timing.loadEventEnd)
```
Robdel12: ❤️ Love this for handling async

src/percy-agent-client/dom.ts
```diff
@@ -202,18 +232,15 @@ class DOM {
...
-    // loop through each form element and apply an ID for serialization later
-    formElements.forEach((elem) => {
+    const frameNodes = this.originalDOM.querySelectorAll('iframe')
+    const elements = [...formNodes, ...frameNodes] as HTMLElement[]
+
+    // loop through each element and apply an ID for serialization later
+    elements.forEach((elem) => {
       if (!elem.getAttribute('data-percy-element-id')) {
-        createUID(elem)
+        elem.setAttribute('data-percy-element-id', createUID())
```
Robdel12: Also, thanks for cleaning up my ugly code 🧹

PR #480, opened 2020-03-06
fix: Log out snapshot options in debug logs

Robdel12:
Super helpful for debugging:

```bash
[percy] -> widths: 375,1200
[percy] -> minHeight: 1024
[percy] -> enableJavaScript: undefined
[percy] -> requestHeaders: undefined
[percy] -> percyCSS: .percy-only-css-global {
  height: 300px;
  width: 300px;
  background-color: blue;
}```

### DIFF
--- a/src/services/agent-service.ts
+++ b/src/services/agent-service.ts
@@ -109,6 +109,12 @@ export class AgentService {
       minHeight: request.body.minHeight || this.snapshotConfig['min-height'],
     }
 
+    snapshotLogger.debug(`-> widths: ${snapshotOptions.widths}`)
+    snapshotLogger.debug(`-> minHeight: ${snapshotOptions.minHeight}`)
+    snapshotLogger.debug(`-> enableJavaScript: ${snapshotOptions.enableJavaScript}`)
+    snapshotLogger.debug(`-> requestHeaders: ${snapshotOptions.requestHeaders}`)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/services/agent-service.ts [hunk 1]
```diff
@@ -109,6 +109,8 @@ export class AgentService {
       minHeight: request.body.minHeight || this.snapshotConfig['min-height'],
     }
 
+    snapshotLogger.debug(`-> snapshot options: ${JSON.stringify(snapshotOptions)}`)
```
wwilsman: Do we want to use JSON or inspect? Just a preference, JSON wraps keys in strings and we use inspect [here](https://github.com/percy/percy-agent/blob/master/src/utils/configuration.ts#L93). Inspect will keep things single line until a certain length, while JSON is all or nothing with the third parameter.

wwilsman:
Or rather than stringify or inspect, we could follow the same pattern from above and break the options apart.

```js
snapshotLogger.debug(`-> widths: ${snapshotOptions.widths}`)
snapshotLogger.debug(`-> minHeight: ${snapshotOptions.minHeight}`)
snapshotLogger.debug(`-> enableJavaScript: ${snapshotOptions.enableJavaScript}`)
snapshotLogger.debug(`-> requestHeaders: ${snapshotOptions.requestHeaders}`)
snapshotLogger.debug(`-> percyCSS: ${snapshotOptions.percyCSS}`)
```

I'm okay with either way, just throwing out options

PR #481, opened 2020-03-09
build: Add Home URL and Source Code URL + Fix

KyleTryon:
Display a link to your home page and the orb source natively in the Orb registry as a clickable link.

Fix: On line 43 - The env_var_name is a string that represents the name of the env var. the "$" is still required to transform it into an env var at the shell.


Robdel12:
Thanks @KyleTryon! :D

PR #483, opened 2020-03-13
feat: Capture Canvas elements

Robdel12:
## What is this?

We can now capture Canvas elements and serialize them into images in the DOM snapshot we capture.

Important to note WebGL canvas elements will need to enable `preserveDrawingBuffer` in order for us to capture them.

### TODOs

- [ ] How to document `preserveDrawingBuffer` is needed? Recipe?


src/percy-agent-client/dom.ts
```diff
@@ -225,6 +226,38 @@ class DOM {
...
+      $image.setAttribute('style', 'max-width: 100%')
+      $image.classList.add('percy-canvas-image')
+
+      $clonedCanvas.style = 'display: none'
+      $image.src = $canvas.toDataURL()
+      $clonedCanvas.setAttribute('data-percy-canvas-serialized', true)
+      $clonedCanvas.parentElement.appendChild($image)
+    }
+
+    this.originalDOM.querySelectorAll('canvas').forEach(canvasToImage)
```
wwilsman: Rather than create a function within a function and `forEach`, can we use a simpler `for` loop?

```js
for (const $canvas of this.originalDOM.querySelectorAll('canvas')) {
  ...
}
```

src/percy-agent-client/dom.ts
```diff
@@ -225,6 +226,38 @@ class DOM {
...
+  private serializeCanvasElements(clonedDOM: HTMLDocument): void {
+    function canvasToImage($canvas: any) {
+      const $image = clonedDOM.createElement('img')
+      const canvasId = $canvas.getAttribute('data-percy-element-id')
+      const $clonedCanvas = clonedDOM.querySelector(`[data-percy-element-id=${canvasId}]`) as any
+
+      $image.setAttribute('style', 'max-width: 100%')
+      $image.classList.add('percy-canvas-image')
+
+      $clonedCanvas.style = 'display: none'
```
wwilsman: Why do we use `style` here for the cloned canvas but `setAttribute` for the image?

src/percy-agent-client/dom.ts
```diff
@@ -225,6 +226,38 @@ class DOM {
...
+   * into the DOM Percy captures.
+   *
+   * It's important to note the `.toDataURL` API requires WebGL canvas elements
+   * to use `preserveDrawingBuffer: true`. This is because `.toDataURL` captures
+   * from the drawing buffer, which is cleared after each render by default for
+   * performance.
+   *
+   */
+  private serializeCanvasElements(clonedDOM: HTMLDocument): void {
+    function canvasToImage($canvas: any) {
```
wwilsman: We probably don't want to do this if JS is enabled, or do we?

Robdel12: OH, excellent question!! I'd say probably not 🤔 The user would need to make sure it can render properly? That's can of worms.

Hm.

src/percy-agent-client/dom.ts
Robdel12: Honestly, how do you feel about deleting the node? This is just carelessly making sure it doesn't somehow appear in snapshots (we don't need it, the image is the source of truth now)

wwilsman: That's true, this is the clone after all, no reason to keep it around

src/percy-agent-client/dom.ts
```diff
@@ -225,6 +226,36 @@ class DOM {
...
+      const $image = clonedDOM.createElement('img')
+      const canvasId = $canvas.getAttribute('data-percy-element-id')
+      const $clonedCanvas = clonedDOM.querySelector(`[data-percy-element-id=${canvasId}]`) as any
+
+      $image.setAttribute('style', 'max-width: 100%')
+      $image.classList.add('percy-canvas-image')
+
+      $image.src = $canvas.toDataURL()
+      $image.setAttribute('data-percy-canvas-serialized', 'true')
+      $clonedCanvas.parentElement.appendChild($image)
```
wwilsman: If the canvas comes first in a container with siblings this will put the image at the end of the container. We should ensure the correct order by inserting it next to the canvas

```suggestion
      $clonedCanvas.parentElement.insertBefore($image, $clonedCanvas)
```

test/integration/agent-integration.test.ts
```diff
@@ -233,5 +233,29 @@ describe('Integration test', () => {
...
+      })
+
+      it("doesn't serialize with JS enabled", async () => {
+        await page.goto(`http://localhost:${PORT}/serialize-canvas.html`)
+        await page.waitFor('#webgl canvas')
+        await page.waitFor(1000)
+        const domSnapshot = await snapshot(page, 'Canvas elements -- with JS', { enableJavaScript: true })
+        const $ = cheerio.load(domSnapshot)
+
+        expect($('[data-percy-canvas-serialized]').length).to.equal(0)
```
wwilsman: Now I'm being nitpicky

```suggestion
        expect($('[data-percy-canvas-serialized]')).to.have.lengthOf(0)
```

test/integration/agent-integration.test.ts
```diff
@@ -233,5 +233,29 @@ describe('Integration test', () => {
...
+      it('captures canvas elements', async () => {
+        await page.goto(`http://localhost:${PORT}/serialize-canvas.html`)
+        await page.waitFor('#webgl canvas')
+        // I cannot think of a nicer way to let the canvas animations/drawing settle
+        // so sadly, use a timeout
+        await page.waitFor(1000)
+        const domSnapshot = await snapshot(page, 'Canvas elements')
+        const $ = cheerio.load(domSnapshot)
+
+        expect($('[data-percy-canvas-serialized]').length).to.equal(2)
```
wwilsman:
```suggestion
        expect($('[data-percy-canvas-serialized]')).to.have.lengthOf(2)
```

PR #507, opened 2020-05-14
fix: Ignore any protocols that aren't `http` or `https`

Robdel12:
## What is this?

The API will throw an error anyways, so we shouldn't bother trying to create resources or snapshots with non `http` or `https` protocols.

### DIFF
--- a/src/services/asset-discovery-service.ts
+++ b/src/services/asset-discovery-service.ts
@@ -175,6 +175,11 @@ export class AssetDiscoveryService extends PercyClientService {
     const parsedRootResourceUrl = new URL(rootResourceUrl)
     const rootUrl = `${parsedRootResourceUrl.protocol}//${parsedRootResourceUrl.host}`
 
+    // Only capture resources with a proper protocol we support capturing
+    if ((/^https?:/).test(resourceUrl)) {
+      return true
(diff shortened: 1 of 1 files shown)


src/services/asset-discovery-service.ts
```diff
@@ -173,8 +173,14 @@ export class AssetDiscoveryService extends PercyClientService {
   // We shouldn't bother passing on requests that will never be saved
   shouldProcessRequest(resourceUrl: string, rootResourceUrl: string): boolean {
     const parsedRootResourceUrl = new URL(rootResourceUrl)
+    const parsedResourceUrl = new URL(resourceUrl)
     const rootUrl = `${parsedRootResourceUrl.protocol}//${parsedRootResourceUrl.host}`
 
+    // Only capture resources with a proper protocol we support capturing
+    if (parsedResourceUrl.protocol === 'http:' || parsedResourceUrl.protocol === 'https:') {
```
wwilsman, line 180: Do you think it's better to parse the URL for the protocol, or use a regexp or string method?

```js
(/^https?:/).test(resourceUrl)
// or
resourceUrl.startsWith('http:') || resourceUrl.startsWith('https:')

PR #510, opened 2020-05-18
fix: Capture resources even when Puppeteer tab timesout

Robdel12:
## What is this?

This splits the try/catches up so we can still capture the discovered resources even if the `goto` to the dom snapshot times out.


Robdel12:
I need a new laptop with a working keyboard ;)

wwilsman:
Excuses excuses, I see your branch name. ;p

PR #512, opened 2020-05-21
fix(🐛): Remove inaccessible JS frames when serializing

wwilsman:
## Purpose

Sometimes ads, trackers, and other external scripts may inject iframes with empty documents and a src of `javascript:false`, `javascript:null`, or `javascript:void`. During serialization, these frames might potentially be skipped and left alone if their (empty) documents are not accessible at the time of serialization. This is a problem for asset discovery when JS is disabled because the `src` attribute is still requested, however, not intercepted or resolved since JS is disabled. This causes the asset discovery browser to hang until the `page.goto` method times out.

## Approach

When the contents of the frame is not accessible in the previous if statement, JS is not enabled, and the frame in question was built with JS, the cloned frame is removed from the cloned DOM.

### Potential pitfalls

Frames that are built with JS by the user might still have these `javascript:*` src attributes. In these cases, the frames will be removed if the document is not accessible. To prevent removal, the frame should be made accessible so that it can be properly serialized.

### DIFF
--- a/src/percy-agent-client/dom.ts
+++ b/src/percy-agent-client/dom.ts
@@ -158,9 +158,12 @@ class DOM {
   }
 
   private serializeFrameElements(clonedDOM: HTMLDocument) {
+    const { enableJavaScript } = this.options
+
     for (const frame of this.originalDOM.querySelectorAll('iframe')) {
       const percyElementId = frame.getAttribute('data-percy-element-id')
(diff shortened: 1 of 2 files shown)


wwilsman:
To expand on "the frame should be made accessible" and be more specific: To allow serialization of JS created frames, the sandbox attribute should contain "allow-same-origin" or there should be no sandbox attribute at all for frames created by JS with a `javascript:*` src attribute.

PR #516, opened 2020-05-29
fix(📌): Pin Puppeteer to 3.1.x

wwilsman:
## Purpose

Puppeteer 3.2 introduces breaking changes by removing the EventEmitter API and not backporting all EventEmitter methods. This update causes agent to error because `page.removeAllListeners` is no longer a method.

## Approach

Pinning to 3.1 is a temporary fix until the breaking changes can be accounted for or we migrate away from Puppeteer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "image-size": "^0.8.2",
     "js-yaml": "^3.13.1",
     "percy-client": "^3.2.0",
-    "puppeteer": "^3.0.2",
+    "puppeteer": "~3.1.0",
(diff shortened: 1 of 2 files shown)


swissspidy:
This was reverted in Puppeteer 3.3.0. Perhaps worth updating?

wwilsman:
Hey @swissspidy! Thanks for pointing out the new release! I was hoping it'd be released yesterday, but I'll get a new version of agent out asap today to unpin puppeteer. 👍

PR #555, opened 2020-08-28
fix: Remove `loading` attributes from cloned iframes

Robdel12:
## What is this?

This attribute causes asset discovery to hang and eventually timeout since we don't scroll the page. Removing the attribute is ideally what we want (to load the iframe).

### DIFF
--- a/src/percy-agent-client/dom.ts
+++ b/src/percy-agent-client/dom.ts
@@ -169,7 +169,6 @@ class DOM {
       // rerendered and do not effect the visuals of a page
       if (clonedDOM.head.contains(cloned)) {
         cloned!.remove()
-
       // if the frame document is accessible, we can serialize it
       } else if (frame.contentDocument) {
         // js is enabled and this frame was built with js, don't serialize it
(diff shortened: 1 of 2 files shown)


src/percy-agent-client/dom.ts
```diff
@@ -169,7 +169,9 @@ class DOM {
       // rerendered and do not effect the visuals of a page
       if (clonedDOM.head.contains(cloned)) {
         cloned!.remove()
-
+      // Remove any lazy loading attributes, since they hang asset discovery
+      } else if (cloned!.getAttribute('loading')) {
```
wwilsman, line 173: This will prevent the following conditionals from being encountered, thus not being able to serialize accessible frames that have this attribute. We should pop this out into it's own block before or after all of these other conditionals.

Robdel12: 🤦 Duh

src/percy-agent-client/dom.ts
```diff
@@ -188,6 +187,11 @@ class DOM {
       } else if (!enableJavaScript && builtWithJs) {
         cloned!.remove()
       }
+
+      // Remove any lazy loading attributes, since they hang asset discovery
+      if (cloned!.getAttribute('loading')) {
```
wwilsman, line 192: I just checked, and `removeAttribute` actually won't produce an error if the attribute doesn't exist. So I think it's safe to avoid the conditional and always remove the loading attribute

PR #562, opened 2020-09-09
fix(🐛): Handle page and snapshot errors

wwilsman:
## Purpose

Sometimes, due to various reasons, the running browser might disconnect from Puppeteer during asset discovery causing a `target closed` error. This error can happen anytime CDP sends a message to the browser, such as enabling JS or setting viewport options. Since it is unhandled it can result in the local server being unable to respond to requests which can then result in request timeouts.

## Approach

Within asset discovery, move any interaction with the `page` object into an error handler. The various request event registrations do not have to be protected since they are generic event emitter registration methods.

I also noticed that none of the methods within the snapshot handler were protected, which might contribute to the server hanging since it does not respond for unhandled async errors by default. I wrapped the snapshot method with another try-catch to help ensure that the server responds when an error is encountered.

### DIFF
--- a/src/services/agent-service.ts
+++ b/src/services/agent-service.ts
@@ -7,7 +7,7 @@ import * as path from 'path'
 import { Configuration } from '../configuration/configuration'
 import { SnapshotConfiguration } from '../configuration/snapshot-configuration'
 import { SnapshotOptions } from '../percy-agent-client/snapshot-options'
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁 LGTM (much needed, I think we can assume none of the promises are caught)

PR #572, opened 2020-10-07
fix(📌): Unpin puppeteer

wwilsman:
## Purpose

In #539, Puppeteer was pinned to a version that was known not to cause browser crashes. Turns out, that that version also crashes, but only in Docker containers where the default shared memory size is too small to properly run Chromium.

Puppeteer passes along the `--disable-dev-shm-usage` flag to Chromium to avoid this issue, but the current pinned Puppeteer version uses Chromium 83 which has a bug that causes it to ignore that flag. https://bugs.chromium.org/p/chromium/issues/detail?id=1085829

## Approach

Newer versions of Puppeteer have come out since we pinned it. This PR unpins it again to use the latest version which seems to avoid other issues outlined in #539 such as missing woff assets.

🤞 

---

This also bumps the version so we can quickly release this soon after merging

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@percy/agent",
   "description": "An agent process for integrating with Percy.",
-  "version": "0.28.2",
+  "version": "0.28.3",
   "author": "Perceptual Inc",
   "bin": {
     "percy": "./bin/run"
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
Pending test fails 🏁 

Oh puppeteer

PR #573, opened 2020-10-13
fix: `toString` PID number for `writeFileSync` (second arg requires string)

Robdel12:
## What is this?

Node 14 will no longer coerce incorrect types given to `writeFileSync`.

image

The error thrown:

```
TypeError: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received type number
```


wwilsman (approved):
🎮

PR #581, opened 2020-11-18
fix: Don't check if the service is running to stop Percy

Robdel12:
## What is this?

This approach assumes running `percy stop` will happen in the same session. The better approach is to send the POST request anyways & catch if it fails. That way if the server is running (on that path/port) it will get stopped.


wwilsman (approved):
🎮

PR #589, opened 2020-12-03
fix(🐛): TypeError: Invalid attempt to spread non-iterable instance on Internet explorer 11

KevinBon:
## Purpose

When using PercyAgent against IE11, not wrapping `querySelectorAll` with `Array.from` will result with `TypeError: Invalid attempt to spread non-iterable instance`.

## More information

I haven't encountered the bug on Chrome, Firefox nor Safari (I haven't tried with Edge).

[Babel related issue](https://github.com/babel/babel/issues/9993)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,13 +132,13 @@ $ yarn test
 You can test Percy Agent Client with
 
 ```bash
-$ yarn test-client
+$ yarn test:client
 ```
 
 And the integration tests can be run with:
 
 ```bash
-$ PERCY_TOKEN=xxx yarn test-integration
+$ PERCY_TOKEN=xxx yarn test:integration
 ```
 
 You can find `PERCY_TOKEN` on your Percy project settings.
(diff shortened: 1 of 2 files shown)


wwilsman:
Hi @KevinBon! Thanks for the PR!

We're hesitant to merge this since this is only necessary for IE11, which Percy does not support and will be end-of-life in 9 months time. While it is a small change that shouldn't have much impact on perf in other browsers, we don't want to encourage capturing DOM from unsupported browsers due to browser inconsistencies and how that DOM is re-rendered in Percy's own browsers.

If you really want to capture DOM from IE11 to render in our other browsers, you might consider polyfilling the missing iterator method on the NodeList: `NodeList.prototype[Symbol.iterator] = Array.prototype[Symbol.iterator];`

KevinBon:
I understand, thank you for taking a look at it @wwilsman!

Robdel12:
Also, FWIW, if/when Percy supports IE 11 for screenshots we'll absolutely revisit :)

PR #611, opened 2021-04-29
build(deps): upgrade puppeteer to v9.0.0

watson:
TypeScript typings are included in this version, so @types/puppeteer has been removed


Robdel12:
Hey @watson! This repo is pretty much deprecated -- we have one more SDK to move off of percy agent to `@percy/cli` before it's fully deprecated. You'll want to check this doc for your SDK and the upgrade path: https://docs.percy.io/docs/migrating-to-percy-cli

PR #646, opened 2022-01-10
fix: Lock colors.js to prevent infinite loop

convenient:
It seems that the lock file has it at 1.4.0 from this commit https://github.com/percy/percy-agent/commit/27127097c8b821fb175c9c8e2e49e31f5a79284c#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR2876

The later versions of colors have an infinite loop added, see references below

References
- https://github.com/Marak/colors.js/commit/074a0f8ed0c31c35d13d28632bd8a049ff136fb6?s=09#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R18
- https://www.theverge.com/2022/1/9/22874949/developer-corrupts-open-source-libraries-projects-affected

Locking this to a known good version should give us capability back while migrations to the newer SDK are still underway

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@oclif/plugin-not-found": "^1.2",
     "axios": "^0.21.1",
     "body-parser": "^1.18.3",
-    "colors": "^1.3.2",
+    "colors": "1.4.0",
     "cors": "^2.8.4",
     "cosmiconfig": "^5.2.1",
     "cross-spawn": "^7.0.2",


Robdel12:
Hey @convenient! I'll give a shot releasing this. But this project has been inactive for a year now, so I can't promise it'll be able to release or new bugs introduced.

I super strongly suggest upgrading, there are bugs that you currently have you probably don't know about. _Ideally,_ upgrading should be fairly straight forward, especially with `@percy/migrate`: https://docs.percy.io/docs/migrating-to-percy-cli (but I know, software always has gremlins)

convenient:
@Robdel12 thanks! Yeah this is a stop gap for some projects running an old version of node that are in the process of being upgraded. 

Those older node versions don't seem to play well with the new SDK so we were using this to keep us going in the meantime.

If we can stop this looping we'll be able to continue using Percy while we work out the migration process and the additional dependencies on our side

convenient:
@Robdel12 🙌  Thank you so much buddy. Real life saver on this one 😅 

This should keep us going while the upgrades are in flight

**Yesterday**
```
$ npx percy exec -- true | head -3
LIBERTY LIBERTY LIBERTY
LIBERTY LIBERTY LIBERTY
LIBERTY LIBERTY LIBERTY
```

**Today**
```
$ npx percy exec -- true | head -3
 ›   Warning: Skipping visual tests. PERCY_TOKEN was not provided.
```

